PR #1, opened 2022-09-23
[FLINK-25756] [connectors/opensearch] Dedicated Opensearch connectors

reta:
Signed-off-by: Andriy Redko <andriy.redko@aiven.io>

## What is the purpose of the change

The goal of this change is to provide dedicated Opensearch connectors [1], [2], [3], [4].

## Brief change log

The implementation is largely based on the existing Elasticsearch 7 connector with a few notable changes (besides the dependencies and APIs):
 - any mentions and uses of mapping types have been removed: it is deprecated feature, scheduled for removal (the indices with mapping types cannot be created or migrated to Opensearch 1.x and beyond)
 - any mentions and uses have been removed: it is deprecated feature, scheduled for removal (only `HighLevelRestClient` is used)
 - the default distributions of Opensearch come with HTTPS turned on, using self-signed certificates: to simplify the integration a new option `allow-insecure` has been added to suppress certificates validation for development and testing purposes
 - old streaming APIs are also supported to facilitate the migration of existing applications from Elasticsearch 7/6 to Opensearch (the classes will change but the familiar model will stay)

The new connector name is `opensearch` and it follows the existing conventions:

```
CREATE TABLE users ( ... ) WITH (
  'connector' = 'opensearch', 
  'hosts' = 'https://localhost:9200',
  'index' = 'users', 
  'allow-insecure' = 'true', 
  'username' = 'admin', 
  'password' = 'admin');
```

## Verifying this change

This change added comprehensive tests and can be verified as follows (largely ported the existing unit and integration tests for Elasticsearch 7):
  - Added unit tests
  - Added integration tests for end-to-end
  - Added end-to-end tests
  - Manually verified the connector by running a node clusters

## Does this pull request potentially affect one of the following parts:

  - Dependencies: yes (the latest Opensearch 1.2.4 APIs as of this moment)
  - The public API, i.e., is any changed class annotated with `@Public(Evolving)`: yes
  - The serializers: no
  - The runtime per-record code paths (performance sensitive): no
  - Anything that affects deployment or recovery: JobManager (and its components), Checkpointing, Kubernetes/Yarn, ZooKeeper: no
  - The S3 file system connector: no

## Documentation

  - Does this pull request introduce a new feature? yes
  - If yes, how is the feature documented? (docs - in progress, JavaDocs)

Huge thanks @snuyanzin for help.
Retargeting https://github.com/apache/flink/pull/18541 to separate repository

[1] https://cwiki.apache.org/confluence/display/FLINK/FLIP-243%3A+Dedicated+Opensearch+connectors
[2] https://www.mail-archive.com/dev@flink.apache.org/msg58911.html
[3] https://lists.apache.org/thread/jls0vqc7jb84jp14j4jok1pqfgo2cl30
[4] https://lists.apache.org/thread/4bms24983g38q956rp8qmm4bpdo4361s


reta:
@MartijnVisser could you please take a look? thank you :-)

reta:
@MartijnVisser doing my deligance by pinging once per month :-)

reta:
@MartijnVisser the POMs and CI scripts have been updated accordingly

flink-connector-opensearch-e2e-tests/pom.xml
```diff
@@ -0,0 +1,104 @@
...
+		 xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
+		 xsi:schemaLocation="http://maven.apache.org/POM/4.0.0 http://maven.apache.org/xsd/maven-4.0.0.xsd">
+
+	<modelVersion>4.0.0</modelVersion>
+
+	<parent>
+		<groupId>org.apache.flink</groupId>
+		<artifactId>flink-connector-opensearch-parent</artifactId>
+		<version>1.0.0-SNAPSHOT</version>
+		<relativePath>..</relativePath>
```
zentol, line 30:
```suggestion
```

flink-connector-opensearch-e2e-tests/pom.xml
```diff
@@ -0,0 +1,104 @@
...
+				<configuration>
+					<artifactItems>
+						<artifactItem>
+							<groupId>org.apache.flink</groupId>
+							<artifactId>flink-connector-test-utils</artifactId>
+							<version>${flink.version}</version>
+							<destFileName>flink-connector-test-utils.jar</destFileName>
+							<type>jar</type>
+							<outputDirectory>${project.build.directory}/dependencies
+							</outputDirectory>
```
zentol, line 96:
```suggestion
							<outputDirectory>${project.build.directory}/dependencies</outputDirectory>
```

flink-connector-opensearch-e2e-tests/pom.xml
```diff
@@ -0,0 +1,104 @@
...
+				<executions>
+					<execution>
+						<id>copy</id>
+						<phase>pre-integration-test</phase>
+						<goals>
+							<goal>copy</goal>
+						</goals>
+					</execution>
+				</executions>
+				<configuration>
```
zentol, line 87: can we move this configuration into the execution?

flink-connector-opensearch/pom.xml
```diff
@@ -0,0 +1,185 @@
...
+	</parent>
+
+	<artifactId>flink-connector-opensearch</artifactId>
+	<name>Flink : Connectors : Opensearch</name>
+
+	<packaging>jar</packaging>
+
+	<!-- Allow users to pass custom connector versions -->
+	<properties>
+		<opensearch.version>1.3.0</opensearch.version>
```
zentol, line 40: Can you give us some background of the opensearch versioning?
I see 2.3.0 artifacts on maven central.

flink-connector-opensearch/pom.xml
```diff
@@ -0,0 +1,185 @@
...
+
+	<dependencies>
+
+		<!-- Core -->
+
+		<dependency>
+			<groupId>org.apache.flink</groupId>
+			<artifactId>flink-connector-base</artifactId>
+			<version>${flink.version}</version>
+		</dependency>
```
zentol, line 51: best to set this to provided to avoid compatibility issues between connector-base and Flink.

flink-connector-opensearch/pom.xml
```diff
@@ -0,0 +1,185 @@
...
+			xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
+			xsi:schemaLocation="http://maven.apache.org/POM/4.0.0 http://maven.apache.org/maven-v4_0_0.xsd">
+
+	<modelVersion>4.0.0</modelVersion>
+
+	<parent>
+		<groupId>org.apache.flink</groupId>
+		<artifactId>flink-connector-opensearch-parent</artifactId>
+		<version>1.0.0-SNAPSHOT</version>
+		<relativePath>..</relativePath>
```
zentol, line 30:
```suggestion
```

flink-connector-opensearch/pom.xml
```diff
@@ -0,0 +1,185 @@
...
+
+	<build>
+		<plugins>
+			<plugin>
+				<groupId>org.apache.maven.plugins</groupId>
+				<artifactId>maven-surefire-plugin</artifactId>
+				<configuration>
+					<!-- Enforce single fork execution because of spawning
+					 Opensearch cluster multiple times -->
+					<forkCount>1</forkCount>
```
zentol, line 179: IIRC we got rid of this in the elasticsearch connector, so we shouldn't need this.

flink-sql-connector-opensearch/pom.xml
```diff
@@ -0,0 +1,159 @@
...
+		 xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
+		 xsi:schemaLocation="http://maven.apache.org/POM/4.0.0 http://maven.apache.org/maven-v4_0_0.xsd">
+
+	<modelVersion>4.0.0</modelVersion>
+
+	<parent>
+		<groupId>org.apache.flink</groupId>
+		<artifactId>flink-connector-opensearch-parent</artifactId>
+		<version>1.0.0-SNAPSHOT</version>
+		<relativePath>..</relativePath>
```
zentol, line 30:
```suggestion
```

flink-sql-connector-opensearch/pom.xml
```diff
@@ -0,0 +1,159 @@
...
+									<artifact>org.apache.lucene:lucene-analyzers-common</artifact>
+									<excludes>
+										<exclude>org/tartarus/**</exclude>
+									</excludes>
+								</filter>
+								<filter>
+									<artifact>*:*</artifact>
+									<excludes>
+										<!-- exclude Java 9 specific classes as otherwise the shade-plugin crashes -->
+										<exclude>META-INF/versions/**</exclude>
```
zentol, line 109: This seems odd; Flink itself is bundling java version-specific files.

Can you show me the error message you got?

flink-sql-connector-opensearch/pom.xml
```diff
@@ -0,0 +1,159 @@
...
+										<exclude>META-INF/services/com.fasterxml.**</exclude>
+										<exclude>META-INF/services/org.apache.lucene.**</exclude>
+										<exclude>META-INF/services/org.opensearch.**</exclude>
+										<exclude>META-INF/LICENSE.txt</exclude>
+									</excludes>
+								</filter>
+							</filters>
+							<relocations>
+								<!-- Force relocation of all Opensearch dependencies. -->
```
zentol, line 118:
```suggestion
```

flink-sql-connector-opensearch/src/main/resources/META-INF/NOTICE
```diff
@@ -0,0 +1,50 @@
+flink-sql-connector-opensearch
+Copyright 2014-2021 The Apache Software Foundation
```
zentol, line 2:
```suggestion
Copyright 2014-2022 The Apache Software Foundation
```

zentol (changes requested):
Made a pass over infrastructure/build/legal files.

reta:
> Can you split the commit such that 1 adds the workflows and another the rest? We can then merge the workflow commit first and actually get CI in this PR.

Sure, https://github.com/apache/flink-connector-opensearch/pull/2, thanks @zentol

flink-connector-opensearch-e2e-tests/pom.xml
reta: > can we move this configuration into the execution?

Sure, may need need same adjustment here https://github.com/apache/flink-connector-elasticsearch/blob/main/flink-connector-elasticsearch-e2e-tests/flink-connector-elasticsearch7-e2e-tests/pom.xml#L109 and https://github.com/apache/flink-connector-elasticsearch/blob/main/flink-connector-elasticsearch-e2e-tests/flink-connector-elasticsearch6-e2e-tests/pom.xml#L108

flink-connector-opensearch/pom.xml
reta: > Can you give us some background of the opensearch versioning? I see 2.3.0 artifacts on maven central.

Sure, so we have 2 maintainable release trains: 1.x and 2.x. The major difference between two (leaving all features aside) is that 1.x has JDK-8 baseline while 2.x has JDK-11 baseline. Since Flink still has JDK-8 baseline, we are using 1.x client BUT it will work with 2.x servers.

flink-connector-opensearch/pom.xml
reta: > best to set this to provided to avoid compatibility issues between connector-base and Flink.

Needs adjustment here https://github.com/apache/flink-connector-elasticsearch/blob/main/flink-connector-elasticsearch-base/pom.xml#L47

flink-sql-connector-opensearch/pom.xml
reta: This one is taken from https://github.com/apache/flink-connector-elasticsearch/blob/main/flink-sql-connector-elasticsearch7/pom.xml#L117 since OpenSearch is a fork of Elasticseach 7.10.2

flink-connector-opensearch-e2e-tests/pom.xml
```diff
@@ -0,0 +1,190 @@
...
+						</configuration>
+					</execution>
+				</executions>
+			</plugin>
+		</plugins>
+	</build>
+
+	<profiles>
+		<profile>
+			<id>run-end-to-end-tests</id>
```
zentol, line 160: There is no longer a real benefit in not treating the e2e tests as normal IT Cases.
(notably because we always run them on CI and running them locally works just fine as well)

flink-connector-opensearch-e2e-tests/pom.xml
```diff
@@ -0,0 +1,190 @@
...
+								<phase>integration-test</phase>
+								<goals>
+									<goal>test</goal>
+								</goals>
+								<configuration>
+									<includes>
+										<include>**/*.*</include>
+									</includes>
+									<!-- E2E tests must not access flink-dist concurrently. -->
+									<forkCount>1</forkCount>
```
zentol, line 178: This shouldn't be required anymore.

flink-connector-opensearch-e2e-tests/pom.xml
```diff
@@ -0,0 +1,190 @@
...
+							<filters combine.children="append">
+								<filter>
+									<artifact>*:*</artifact>
+									<excludes>
+										<exclude>META-INF/*.SF</exclude>
+										<exclude>META-INF/*.DSA</exclude>
+										<exclude>META-INF/*.RSA</exclude>
+									</excludes>
+								</filter>
+							</filters>
```
zentol, line 125: If you'd set execution id to `shade-flink` this wouldn't be necessary.

pom.xml
```diff
@@ -327,6 +327,12 @@ under the License.
 			</dependency>
 
 			<!-- For dependency convergence -->
+			<dependency>
+				<groupId>net.java.dev.jna</groupId>
+				<artifactId>jna</artifactId>
```
zentol, line 332: Where this come from?

flink-sql-connector-opensearch/pom.xml
zentol: What happens without it? :see_no_evil:

flink-connector-opensearch-e2e-tests/pom.xml
reta: @zentol this it not correct, the e2e tests need Flink distribution to be downloaded and provided, this preparation is done within CI workflows, otherwise the tests won't even start.

pom.xml
reta: Dependency convergence, the OpenSearch brings older version

flink-sql-connector-opensearch/pom.xml
reta: Could be quite outdated, removed (for now)

flink-connector-opensearch-e2e-tests/pom.xml
zentol: That's a good point.

We should fix that though (in a follow-up; across connectors); the tests should be able to download Flink themselves. These magical prerequisite steps just make it difficult to test things locally.

flink-connector-opensearch-e2e-tests/pom.xml
reta: It seems like even `shade-flink` is not needed, removed all that

flink-connector-opensearch-e2e-tests/pom.xml
reta: :+1: second that. It is, took me a while to understand how I could run it locally

flink-sql-connector-opensearch/pom.xml
reta: > What happens without it? see_no_evil

@zentol answer to your (and mine) curiosity:

```
 Error:  Failed to execute goal org.apache.maven.plugins:maven-shade-plugin:3.1.1:shade (shade-flink) on project flink-sql-connector-opensearch: Error creating shaded jar: Problem shading JAR /home/runner/.m2/repository/org/opensearch/opensearch/1.3.0/opensearch-1.3.0.jar entry META-INF/versions/11/org/opensearch/monitor/jvm/JvmPid.class: java.lang.IllegalArgumentException -> [Help 1]
``

Comment (zentol) [reply flink-sql-connector-opensearch/pom.xml:108]: Alright then. :sweat_smile:

File: pom.xml (hunk 15 above)
Comment (zentol) [reply pom.xml:332]: So could we exclude it in general? Or do we need it for some tests to work?

What is the bootstrap licensing issue about?

File: flink-connector-opensearch/src/main/java/org/apache/flink/connector/opensearch/sink/OpensearchSinkBuilder.java [hunk 16]
```diff
@@ -0,0 +1,449 @@
...
+     *
+     * @param allowInsecure allow or not to insecure network endpoints
+     * @return this builder
+     */
+    public OpensearchSinkBuilder<IN> setAllowInsecure(boolean allowInsecure) {
+        this.allowInsecure = allowInsecure;
+        return self();
+    }
+
+    protected BulkProcessorBuilderFactory getBulkProcessorBuilderFactory() {
```
zentol, line 300: There's no reason for this to be protected since BulkProcessorBuilderFactory is an internal interface. 
Maybe that entire class shouldn't exist in it's current form, as it only existed to supported different elasticsearch versions (unless we intend to do the same in the near future for opensearch).

pom.xml
```diff
@@ -0,0 +1,463 @@
...
+		<url>https://github.com/apache/flink-connector-opensearch</url>
+		<connection>git@github.com:apache/flink-connector-opensearch.git</connection>
+		<developerConnection>
+			scm:git:https://gitbox.apache.org/repos/asf/flink-connector-opensearch.git
+		</developerConnection>
+	</scm>
+
+	<modules>
+		<module>flink-connector-opensearch</module>
+		<module>flink-connector-opensearch-e2e-tests</module>
```
zentol, line 56: missing the sql connector module

flink-connector-opensearch/src/main/java/org/apache/flink/connector/opensearch/sink/OpensearchSink.java
```diff
@@ -0,0 +1,91 @@
...
+ *       actions might be sent to Opensearch multiple times when Flink restarts. These additional
+ *       requests may cause inconsistent data in Opensearch right after the restart, but eventually
+ *       everything will be consistent again.
+ * </ul>
+ *
+ * @param <IN> type of the records converted to Opensearch actions
+ * @see OpensearchSinkBuilder on how to construct a OpensearchSink
+ */
+@PublicEvolving
+public class OpensearchSink<IN> implements Sink<IN> {
```
zentol, line 54: Curious why this class lost the `getDeliveryGuarantee()`

flink-connector-opensearch/src/main/java/org/apache/flink/connector/opensearch/sink/NetworkClientConfig.java
```diff
@@ -0,0 +1,86 @@
...
+
+    @Nullable
+    public String getConnectionPathPrefix() {
+        return connectionPathPrefix;
+    }
+
+    @Nullable
+    public Boolean isAllowInsecure() {
+        return allowInsecure;
+    }
```
zentol, line 85: Consider improving on the existing code and returning an Optional instead.

flink-connector-opensearch/src/main/java/org/apache/flink/connector/opensearch/table/OpensearchConfiguration.java
```diff
@@ -0,0 +1,168 @@
...
+
+    public Optional<Integer> getBulkFlushBackoffRetries() {
+        return config.getOptional(BULK_FLUSH_BACKOFF_MAX_RETRIES_OPTION);
+    }
+
+    public Optional<Long> getBulkFlushBackoffDelay() {
+        return config.getOptional(BULK_FLUSH_BACKOFF_DELAY_OPTION).map(Duration::toMillis);
+    }
+
+    public boolean isDisableFlushOnCheckpoint() {
```
zentol, line 101: unused. I also cant find the corresponding option in the ES connector; where does it come from?

flink-connector-opensearch/src/main/java/org/apache/flink/connector/opensearch/table/OpensearchConnectorOptions.java
```diff
@@ -0,0 +1,161 @@
...
+                    .noDefaultValue()
+                    .withDescription("Maximum number of retries.");
+
+    public static final ConfigOption<Duration> BULK_FLUSH_BACKOFF_DELAY_OPTION =
+            ConfigOptions.key("sink.bulk-flush.backoff.delay")
+                    .durationType()
+                    .noDefaultValue()
+                    .withDescription("Delay between each backoff attempt.");
+
+    public static final ConfigOption<Boolean> FLUSH_ON_CHECKPOINT_OPTION =
```
zentol, line 108: effectively unused

flink-connector-opensearch/src/main/java/org/apache/flink/connector/opensearch/table/OpensearchDynamicSink.java
```diff
@@ -0,0 +1,201 @@
...
+ * A {@link DynamicTableSink} that describes how to create a {@link OpensearchSink} from a logical
+ * description.
+ */
+@Internal
+class OpensearchDynamicSink implements DynamicTableSink {
+
+    final EncodingFormat<SerializationSchema<RowData>> format;
+    final DataType physicalRowDataType;
+    final List<LogicalTypeWithIndex> primaryKeyLogicalTypesWithIndex;
+    final OpensearchConfiguration config;
```
zentol, line 54: The lack of the localTimeZoneId field implies that FLINK-24571 isn't contained in this PR. I'm wondering why that is.

flink-connector-opensearch/src/main/java/org/apache/flink/connector/opensearch/table/OpensearchDynamicSinkFactory.java
```diff
@@ -0,0 +1,222 @@
...
+
+/** A {@link DynamicTableSinkFactory} for discovering OpensearchDynamicSink. */
+@Internal
+public class OpensearchDynamicSinkFactory implements DynamicTableSinkFactory {
+    private static final String FACTORY_IDENTIFIER = "opensearch";
+
+    private final OpensearchSinkBuilderSupplier<RowData> sinkBuilderSupplier;
+
+    public OpensearchDynamicSinkFactory() {
+        this.sinkBuilderSupplier = OpensearchSinkBuilder<RowData>::new;
```
zentol, line 78:
```suggestion
        this.sinkBuilderSupplier = OpensearchSinkBuilder::new;
```
not required

zentol: This field doesn't seem necessary in the first place; we could handle that within the sink.

flink-connector-opensearch/src/main/java/org/apache/flink/connector/opensearch/table/OpensearchDynamicSinkFactory.java
```diff
@@ -0,0 +1,222 @@
...
+        return new OpensearchDynamicSink(
+                format,
+                config,
+                primaryKeyLogicalTypesWithIndex,
+                context.getPhysicalRowDataType(),
+                capitalize(FACTORY_IDENTIFIER),
+                sinkBuilderSupplier);
+    }
+
+    OpensearchConfiguration getConfiguration(Context context) {
```
zentol, line 100: can be private, probably also applies to other methods in here.

flink-connector-opensearch/src/main/java/org/apache/flink/connector/opensearch/table/OpensearchDynamicSinkFactory.java
```diff
@@ -0,0 +1,222 @@
...
+        }
+    }
+
+    static void validate(boolean condition, Supplier<String> message) {
+        if (!condition) {
+            throw new ValidationException(message.get());
+        }
+    }
+
+    EncodingFormat<SerializationSchema<RowData>> getValidatedEncodingFormat(
```
zentol, line 154: Where does this come from?

flink-connector-opensearch/src/main/java/org/apache/flink/connector/opensearch/table/OpensearchDynamicSinkFactory.java
```diff
@@ -0,0 +1,222 @@
...
+                .collect(Collectors.toList());
+    }
+
+    @Override
+    public Set<ConfigOption<?>> requiredOptions() {
+        return Stream.of(HOSTS_OPTION, INDEX_OPTION).collect(Collectors.toSet());
+    }
+
+    @Override
+    public Set<ConfigOption<?>> optionalOptions() {
```
zentol, line 196: Why is this missing `forwardOptions()` (FLINK-25391)?

zentol (commented):
Do you, by chance, have a branch that shows the _actual_ changes made on the ES connector code?
Otherwise I'll just have to continue manually comparing them one-by-one and coming back to you on any differences found.

pom.xml
reta: It is not missing, driven by profile [1] which is active by default. Reason: it comes from Flink itself [2]

[1] https://github.com/apache/flink-connector-opensearch/pull/1/files/41cc7bb681aede94c543c28cc5629802874aac70#diff-9c5fb3d1b7e3b0f54bc5c4182965c4fe1f9023d449017cece3005d3f90e8e4d8R385
[2] https://github.com/apache/flink/blob/master/flink-connectors/pom.xml#L83

flink-connector-opensearch/src/test/java/org/apache/flink/streaming/connectors/opensearch/OpensearchSinkITCase.java
```diff
@@ -0,0 +1,181 @@
...
+        OpensearchSink.Builder<Tuple2<Integer, String>> builder =
+                new OpensearchSink.Builder<>(httpHosts, opensearchSinkFunction);
+        builder.setBulkFlushMaxActions(bulkFlushMaxActions);
+
+        return builder.build();
+    }
+
+    private OpensearchSink<Tuple2<Integer, String>> createOpensearchSinkForNode(
+            int bulkFlushMaxActions,
+            String clusterName,
```
zentol, line 141: clusterName is unused.

pom.xml
reta: Sure, thanks for clarification, removing the profile

flink-connector-opensearch/src/main/java/org/apache/flink/connector/opensearch/sink/OpensearchSink.java
reta: If you mean the equivalent from `ElasticseachSink` :
```
    @VisibleForTesting
    DeliveryGuarantee getDeliveryGuarantee() {
        return deliveryGuarantee;
    }
```

The default for `deliveryGuarantee` was changed recently and the method was added to tests specifically for that.

flink-connector-opensearch/src/main/java/org/apache/flink/streaming/connectors/opensearch/OpensearchSink.java
```diff
@@ -0,0 +1,807 @@
...
+    private static Throwable extractFailureCauseFromBulkItemResponse(
+            BulkItemResponse bulkItemResponse) {
+        if (!bulkItemResponse.isFailed()) {
+            return null;
+        } else {
+            return bulkItemResponse.getFailure().getCause();
+        }
+    }
+
+    long getNumPendingRequests() {
```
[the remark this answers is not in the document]
zentol: Indeed that entire test class appears to be missing.

flink-connector-opensearch/src/main/java/org/apache/flink/connector/opensearch/sink/NetworkClientConfig.java
reta: :+1:

flink-connector-opensearch-e2e-tests/src/test/java/org/apache/flink/streaming/tests/OpensearchSinkE2ECase.java
```diff
@@ -0,0 +1,117 @@
...
+    private static final int READER_TIMEOUT = -1; // Not used
+
+    @TestSemantics
+    CheckpointingMode[] semantics = new CheckpointingMode[] {CheckpointingMode.EXACTLY_ONCE};
+
+    @TestEnv
+    protected FlinkContainerTestEnvironment flink = new FlinkContainerTestEnvironment(1, 6);
+
+    public OpensearchSinkE2ECase() throws Exception {}
+
```
zentol, line 59:
```suggestion
```

flink-connector-opensearch-e2e-tests/src/test/java/org/apache/flink/streaming/tests/OpensearchSinkE2ECase.java
```diff
@@ -0,0 +1,117 @@
...
+public class OpensearchSinkE2ECase extends SinkTestSuiteBase<ComparableTuple2<Integer, String>> {
+    private static final Logger LOG = LoggerFactory.getLogger(OpensearchSinkE2ECase.class);
+    private static final int READER_RETRY_ATTEMPTS = 10;
+    private static final int READER_TIMEOUT = -1; // Not used
+
+    @TestSemantics
+    CheckpointingMode[] semantics = new CheckpointingMode[] {CheckpointingMode.EXACTLY_ONCE};
+
+    @TestEnv
+    protected FlinkContainerTestEnvironment flink = new FlinkContainerTestEnvironment(1, 6);
```
zentol, line 56:
```suggestion
    FlinkContainerTestEnvironment flink = new FlinkContainerTestEnvironment(1, 6);
```

flink-connector-opensearch/src/main/java/org/apache/flink/connector/opensearch/table/OpensearchConfiguration.java
reta: It was there I believe right around the time we started to work on it [1] last year

[1] https://github.com/apache/flink/commit/6609570e49901d83911ec68ced6465f49658c791

flink-connector-opensearch/src/main/java/org/apache/flink/connector/opensearch/table/OpensearchDynamicSinkFactory.java
reta: > This field doesn't seem necessary in the first place; we could handle that within the sink.

Not so sure, the tradeoff is to create supplier instance only once, vs creating it all the time in place

flink-connector-opensearch/src/main/java/org/apache/flink/connector/opensearch/table/OpensearchDynamicSinkFactory.java
reta: Was there before FLINK-25391, changed implementation

flink-connector-opensearch/src/main/java/org/apache/flink/streaming/connectors/opensearch/OpensearchSink.java
```diff
@@ -0,0 +1,807 @@
...
+    }
+
+    /**
+     * Creates an Opensearch client implementing {@link AutoCloseable}.
+     *
+     * @param clientConfig The configuration to use when constructing the client.
+     * @return The created client.
+     * @throws IOException IOException
+     */
+    private RestHighLevelClient createClient(Map<String, String> clientConfig) throws IOException {
```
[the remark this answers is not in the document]
reta: :+1: Removed

flink-connector-opensearch/src/main/java/org/apache/flink/streaming/connectors/opensearch/util/IgnoringFailureHandler.java
```diff
@@ -0,0 +1,38 @@
...
+
+import org.apache.flink.annotation.Internal;
+import org.apache.flink.streaming.connectors.opensearch.ActionRequestFailureHandler;
+import org.apache.flink.streaming.connectors.opensearch.RequestIndexer;
+
+import org.opensearch.action.ActionRequest;
+
+/** Ignores all kinds of failures and drops the affected {@link ActionRequest}. */
+@Internal
+public class IgnoringFailureHandler implements ActionRequestFailureHandler {
```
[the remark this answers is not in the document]
reta: I believe these 2 handlers (IgnoringFailureHandler and RetryRejectedExecutionFailureHandler) are provided as an alternatives for configuring OpensearchSink::setFailureHandler

flink-connector-opensearch/src/main/java/org/apache/flink/streaming/connectors/opensearch/util/RetryRejectedExecutionFailureHandler.java
```diff
@@ -0,0 +1,60 @@
...
+/**
+ * An {@link ActionRequestFailureHandler} that re-adds requests that failed due to temporary {@link
+ * OpenSearchRejectedExecutionException}s (which means that Opensearch node queues are currently
+ * full), and fails for all other failures.
+ *
+ * @deprecated This hase been deprecated and will be removed in the future.
+ */
+@Deprecated
+@PublicEvolving
+public class RetryRejectedExecutionFailureHandler implements ActionRequestFailureHandler {
```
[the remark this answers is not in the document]
reta: Same as above - for users to use

flink-connector-opensearch/src/test/java/org/apache/flink/connector/opensearch/table/IndexGeneratorTest.java
```diff
@@ -0,0 +1,264 @@
...
+    }
+
+    @Test
+    public void testDynamicIndexUnsupportedFormat() {
+        String expectedExceptionMsg = "Unsupported field: HourOfDay";
+        try {
+            IndexGeneratorFactory.createIndexGenerator(
+                    "my-index-{local_date|yyyy/MM/dd HH:mm}", fieldNames, dataTypes);
+        } catch (UnsupportedTemporalTypeException e) {
+            Assertions.assertEquals(expectedExceptionMsg, e.getMessage());
```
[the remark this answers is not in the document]
reta: That's a big one (as a change), will try to tackle it at some point this week

flink-connector-opensearch-e2e-tests/src/test/java/org/apache/flink/streaming/tests/OpensearchSinkE2ECase.java
reta: This is needed, the reason for that is that initialization for fields in this class could throw `Exception` (which is not allowed by default), the default constructor with `throws Exception` spec hints the compiler

flink-connector-opensearch/src/main/java/org/apache/flink/connector/opensearch/table/OpensearchDynamicSinkFactory.java
zentol: If OpensearchDynamicSink#getSinkRuntimeProvider directly creates the opensearch sink builder without going through the supplier then it should be functionally identical.
(You don't need a supplier at all.)

flink-connector-opensearch/src/main/java/org/apache/flink/streaming/connectors/opensearch/util/IgnoringFailureHandler.java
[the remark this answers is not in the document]
zentol: Yes and no. A user _could_ set the failure handler, but this implementation here is marked as `@Internal`; presumably because they were targeting the Table API connector.
See `ElasticsearchConfiguration#getFailureHandler`

_If_ we want to be be a drop-in replacement for the ES table api connector, then we'd need to pull in all of FLINK-26638, which also includes going back to the old ES sink for the Table API.

It's a bit of an unsolved question on the ES connector side as to what will happen with these failure handlers :/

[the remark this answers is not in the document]
reta: Sure, thanks, I will drop them

flink-connector-opensearch/src/main/java/org/apache/flink/connector/opensearch/sink/OpensearchSinkBuilder.java
reta: > There's no reason for this to be protected since BulkProcessorBuilderFactory is an internal interface

:+1: 

> Maybe that entire class shouldn't exist in it's current form, as it only existed to supported different elasticsearch versions (unless we intend to do the same in the near future for opensearch).

Yes, it could be removed since the only place `BulkProcessor` is need is inside the `OpensearchWriter`

reta:
> The architecture tests are missing. Have a look at the ES connector.

Added

flink-connector-opensearch/src/main/java/org/apache/flink/connector/opensearch/sink/OpensearchSink.java
```diff
@@ -88,4 +88,8 @@ public SinkWriter<IN> createWriter(InitContext context) throws IOException {
                 context.metricGroup(),
                 context.getMailboxExecutor());
     }
+
+    DeliveryGuarantee getDeliveryGuarantee() {
```
zentol, line 88: add `@VisibleForTesting`

flink-connector-opensearch/src/main/java/org/apache/flink/connector/opensearch/sink/RequestIndexer.java
```diff
@@ -0,0 +1,56 @@
...
+
+import org.opensearch.action.delete.DeleteRequest;
+import org.opensearch.action.index.IndexRequest;
+import org.opensearch.action.update.UpdateRequest;
+
+/**
+ * Users add multiple delete, index or update requests to a {@link RequestIndexer} to prepare them
+ * for sending to an Opensearch cluster.
+ */
+@Internal
```
zentol, line 31: This shouldn't be internal since it's returned by OpensearchEmitter which is PublicEvolving.
(as reported by the architecture tests ;) )

flink-connector-opensearch/src/test/java/org/apache/flink/architecture/TestCodeArchitectureTest.java
```diff
@@ -0,0 +1,57 @@
...
+
+/** Architecture tests for test code. */
+@AnalyzeClasses(
+        packages = {
+            "org.apache.flink.connector.opensearch",
+            "org.apache.flink.streaming.connectors.opensearch"
+        },
+        importOptions = {
+            ImportOption.OnlyIncludeTests.class,
+            TestCodeArchitectureTest.IncludeES7ImportOption.class,
```
zentol, line 39: This isn't required; it's only needed when you have a `base` module like the ES connector does.

pom.xml
```diff
@@ -0,0 +1,447 @@
...
+
+	<properties>
+		<project.build.sourceEncoding>UTF-8</project.build.sourceEncoding>
+		<project.reporting.outputEncoding>UTF-8</project.reporting.outputEncoding>
+
+		<flink.version>1.16.0</flink.version>
+		<flink.shaded.version>15.0</flink.shaded.version>
+		
+		<jackson-bom.version>2.13.4.20221013</jackson-bom.version>
+		<junit4.version>4.13.2</junit4.version>
```
zentol, line 68: JUnit4 should no longer be required.

flink-connector-opensearch/src/test/java/org/apache/flink/architecture/TestCodeArchitectureTest.java
reta: Ah ... definitely the copy/paste leftovers

flink-connector-opensearch/src/main/java/org/apache/flink/connector/opensearch/sink/OpensearchSink.java
zentol: The architecture tests verify that these methods are only called in tests ;) It allows us to leak some implementation details into tests without impacting production code. Can make your life a lot easier.

reta: I believe in APIs :) I think there are number of ways to approach testing of implementation details, but that's certainly very opinionated subjects, thanks for sharing your point of view.

flink-connector-opensearch/src/main/java/org/apache/flink/streaming/connectors/opensearch/OpensearchSink.java
```diff
@@ -0,0 +1,807 @@
...
+ *
+ * <p>This class implements the common behaviour across Opensearch versions, such as the use of an
+ * internal {@link BulkProcessor} to buffer multiple {@link ActionRequest}s before sending the
+ * requests to the cluster, as well as passing input records to the user provided {@link
+ * OpensearchSinkFunction} for processing.
+ *
+ * @param <T> Type of the elements handled by this sink
+ */
+@Internal
+public class OpensearchSink<T> extends RichSinkFunction<T> implements CheckpointedFunction {
```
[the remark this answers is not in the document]
reta: Similarly to `org.apache.flink.streaming.connectors.elasticsearch*.ElasticsearchSink`? Probably makes sense

reta:
@zentol thnaks a lot for the review, I think I addressed all your comments, please let me know if I missed some, thanks!

> @reta There are some unresolved comments, like [#1 (comment)](https://github.com/apache/flink-connector-opensearch/pull/1#discussion_r1029128244), [#1 (comment)](https://github.com/apache/flink-connector-opensearch/pull/1#discussion_r1029342543), [#1 (comment)](https://github.com/apache/flink-connector-opensearch/pull/1#discussion_r1029129392).

Thanks @zentol , addressed or/and answered

flink-connector-opensearch/src/main/java/org/apache/flink/streaming/connectors/opensearch/OpensearchSink.java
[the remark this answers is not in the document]
reta: > add integration tests instead (to keep the API contained)

This is done now

flink-connector-opensearch/src/test/java/org/apache/flink/streaming/connectors/opensearch/OpensearchSinkTest.java
```diff
@@ -0,0 +1,562 @@
...
+                                new Failure(
+                                        "test",
+                                        "_doc",
+                                        "2",
+                                        new Exception("artificial failure for record")))));
+
+        testHarness.processElement(new StreamRecord<>("msg-1"));
+
+        // Await for flush to be complete
+        awaitForFlushToFinish(1);
```
zentol, line 212: You can avoid this kind of busy waiting quite easily by using `OneShotLatches`.

```
    private OneShotLatch addResponse(Consumer<HttpResponse> consumer) {
        OneShotLatch oneShotLatch = new OneShotLatch();
        responses.add(
                response -> {
                    consumer.accept(response);
                    oneShotLatch.trigger();
                });
        return oneShotLatch;
    }
....
      OneShotLatch firstResponse = responses.add(
                createResponse(
                        new BulkItemResponse(
                                1,
                                OpType.INDEX,
                                new IndexResponse(
                                        new ShardId("test", "-", 0), "_doc", "1", 0, 0, 1, true))));
...
    firstResponse.await();
```

That being said, this whole "wait for server response to be consumed" approach is a bit sketchy?
We're only waiting for the server to send the response, _not_ for the connector to process it.

flink-connector-opensearch/src/test/java/org/apache/flink/streaming/connectors/opensearch/OpensearchSinkTest.java
```diff
@@ -0,0 +1,562 @@
...
+                                new Failure(
+                                        "test",
+                                        "_doc",
+                                        "1",
+                                        new Exception("artificial failure for record")))));
+
+        testHarness.processElement(new StreamRecord<>("msg"));
+
+        // current number of pending request should be 1 due to the re-add
+        assertThat(sink.getNumPendingRequests()).isEqualTo(1);
```
zentol, line 395: This assertion sometimes fails locally when the test is run in a loop.
```
expected: 1L
 but was: 0L
```

flink-connector-opensearch/src/test/java/org/apache/flink/streaming/connectors/opensearch/OpensearchSinkTest.java
```diff
@@ -0,0 +1,562 @@
...
+                };
+        snapshotThread.start();
+
+        // Await for flush to be complete
+        awaitForFlushToFinish(0);
+
+        assertThatThrownBy(snapshotThread::sync)
+                .getCause()
+                .getCause()
+                .hasMessageContaining("Unable to parse response body");
```
zentol, line 354: This sounds like an error in the test setup.

flink-connector-opensearch/src/test/java/org/apache/flink/streaming/connectors/opensearch/OpensearchSinkTest.java
```diff
@@ -0,0 +1,562 @@
...
+
+        @Override
+        public void onFailure(
+                ActionRequest action, Throwable failure, int restStatusCode, RequestIndexer indexer)
+                throws Throwable {
+            indexer.add(action);
+        }
+    }
+
+    private Consumer<HttpResponse> createResponse(BulkItemResponse item) {
```
zentol, line 539: can be static

zentol (changes requested):

flink-connector-opensearch/src/test/java/org/apache/flink/streaming/connectors/opensearch/OpensearchSinkTest.java
reta: Aha, I think I know why, will fix that

flink-connector-opensearch/src/test/java/org/apache/flink/streaming/connectors/opensearch/OpensearchSinkTest.java
reta: Not really, so the original test case is designed to verify the condition "complete bulk response fails". The Bulk API (in Elasticsearch or Opensearch) has does not support that, the failure are only part of individual bulk item [1]. The way I was able to replicate this kind of scenario is by sending invalid payload as part of bulk response but I could not control what kind of exception should be raised (hence we have `Unable to parse response body`), I could add comment about it to clarify it is by design.

[1] https://www.elastic.co/guide/en/elasticsearch/reference/current/docs-bulk.html#docs-bulk-api-example

flink-connector-opensearch/src/test/java/org/apache/flink/streaming/connectors/opensearch/OpensearchSinkTest.java
reta: I would argue there is definitely no need for additional latch: the queue is the "latch" by itself - we know when it is drained. We could also use https://github.com/awaitility/awaitility to make the waiting logic clean.

> That being said, this whole "wait for server response to be consumed" approach is a bit sketchy?

We are dealing with async system so I think this is kind of expectable: we are waiting for some events to happen at some point (be it latch, queue or other means). The sink does not provide any means to hook inside and, for example, we could observe the "process it" part using `getNumPendingRequests` only.

flink-connector-opensearch/src/test/java/org/apache/flink/streaming/connectors/opensearch/OpensearchSinkTest.java
zentol: > The Bulk API (in Elasticsearch or Opensearch) does not support that

Couldn't it fail as a whole because of connection issues etc.?

In any case this should be fine then.

flink-connector-opensearch/src/test/java/org/apache/flink/streaming/connectors/opensearch/OpensearchSinkTest.java
reta: > The point was to not wait by looping over a condition to save CI resources. 

Sure, replaced with `Condition`, should be simpler than per-response monitors.

> This is especially a problem when the condition is never fulfilled and we just churn through CPU cycles for no good reason.

All tests are timeboxed to 5 seconds

flink-connector-opensearch/src/test/java/org/apache/flink/streaming/connectors/opensearch/OpensearchSinkTest.java
```diff
@@ -0,0 +1,586 @@
...
+                }
+                response.setEntity(
+                        new ByteArrayEntity(baos.toByteArray(), ContentType.APPLICATION_JSON));
+            } catch (final IOException ex) {
+                response.setStatusCode(500);
+            }
+        };
+    }
+
+    private static void awaitForCondition(Supplier<Boolean> condition) throws InterruptedException {
```
reta, line 572: There is only 1 place when we need that, due to the fact that `failureRequestIndexer` is not immediately re-adding the requests but does it on the next batch interval only

flink-connector-opensearch/src/test/java/org/apache/flink/streaming/connectors/opensearch/OpensearchSinkTest.java
zentol: > All tests are timeboxed to 5 seconds

a) In our experience such small timeouts actually cause tests to be unstable.
b) code has a tendency to travel. Arguing that "this code is fine because of that code _over there_" basically means "this code is unsafe to copy", which isn't a good state to be in.

flink-connector-opensearch/src/test/java/org/apache/flink/streaming/connectors/opensearch/OpensearchSinkTest.java
```diff
@@ -79,7 +82,15 @@ public void setUp() throws IOException {
                         return (req, resp, context) -> resp.setStatusCode(200);
                     } else if (method.equalsIgnoreCase("POST")) {
                         // Bulk responses are configured per test case
-                        return (req, resp, context) -> responses.poll().accept(resp);
+                        return (req, resp, context) -> {
+                            lock.lock();
+                            try {
+                                responses.poll().accept(resp);
+                                flushed.signalAll();
```
zentol, line 89: Or you could just use a OneShotLatch which makes all of this logic here a one-liner _and_ gives you more control to wait for a specific message to be consumed :shrug: 
(and you wouldn't need `awaitForCondition`)

reta: `awaitForCondition` is not checking the flushing part, it is checking the `numPendingRequest`, which will be eventually updated

flink-connector-opensearch/src/test/java/org/apache/flink/streaming/connectors/opensearch/OpensearchSinkTest.java
reta: The tests seem to be very stable, I was running them continuously for 2h and have not seen any flakyness.

zentol: It's more of a CI problem. Time jumping ahead and things like that. :weary:

reta: I think we could always update the test cases, if the instability comes in, the 5 seconds buffer is 3x buffer for what test case should take.

reta:
@zentol thanks on more time for the review, I think none of your comments left unattended, thanks

wbeckler:
It would be amazing if this could be based on the opensearch-java client rather than RHLC, since that is where all of the improvements are going. Is that how this is set up?

reta:
> It would be amazing if this could be based on the opensearch-java client rather than RHLC, since that is where all of the improvements are going. Is that how this is set up?

Not for now, see please https://github.com/opensearch-project/opensearch-java/issues/181 (this is what the connector uses).

pom.xml
```diff
@@ -0,0 +1,440 @@
...
+				<exclusions>
+					<exclusion>
+						<groupId>log4j</groupId>
+						<artifactId>log4j</artifactId>
+					</exclusion>
+					<exclusion>
+						<groupId>org.slf4j</groupId>
+						<artifactId>slf4j-log4j12</artifactId>
+					</exclusion>
+				</exclusions>
```
zentol, line 209:
```suggestion
```
Since 1.16.0 this should no longer be required.

zentol (approved):

zentol (changes requested):
Just noticed that we're lacking documentation; like https://github.com/apache/flink-connector-elasticsearch/blob/main/docs/content/docs/connectors/datastream/elasticsearch.md.

reta:
> Just noticed that we're lacking documentation; like https://github.com/apache/flink-connector-elasticsearch/blob/main/docs/content/docs/connectors/datastream/elasticsearch.md.

Thanks @zentol, I have it ready and thought to open separate pull request (to split MDs from code), makes sense to you?

zentol:
ok

ypark2103:
hi @zentol @reta 
I see that flink-opensearch-connector was merged to master recently but I don't see it was registered in maven repository.
When will this be added to maven repository so I can use this from pom.xml? I also don't see opensearch connector in official flink documentation. Can this library be used in Java 8 also?

reta:
> hi @zentol @reta I see that [flink-opensearch-connector](https://issues.apache.org/jira/browse/FLINK-opensearch-connector) was merged to master recently but I don't see it was registered in maven repository. When will this be added to maven repository so I can use this from pom.xml? 

Hey @ypark2103 , the vote was ongoing and just approved [1], expect artifacts to be available in the coming days.

[1] https://lists.apache.org/list?dev@flink.apache.org:lte=1M:Opensearch

> I also don't see opensearch connector in official flink documentation. 

The Opensearch documentations available here [2] but I am not sure how it is going to be integrated into Flink documentation, @zentol may be you could spot some light?

[2] https://github.com/apache/flink-connector-opensearch/tree/main/docs

> Can this library be used in Java 8 also?

Yes

ypark2103:
@reta Is there a way we can install this library with pom.xml?
I don't see opensearch in maven depository yet like elasticsearch connector.
https://mvnrepository.com/artifact/org.apache.flink/flink-connector-elasticsearch7

reta:
> The artifacts you mentioned that will be available in the coming days is referring to registering this library to maven repo?

@ypark2103 Correct

MartijnVisser:
> The Opensearch documentations available here [2] but I am not sure how it is going to be integrated into Flink documentation, @zentol may be you could spot some light?

They will be integrated via https://github.com/apache/flink/pull/21518 when the final steps for the release process have been completed :)

ypark2103:
@reta @MartijnVisser Is this Flink-Connector-Opensearch only available for flink 1.16 version? What about flink 1.11 or 1.12?

reta:
@ypark2103 that is correct, the Flink Opensearch Connector was following the Flink's externalization model for connectors, the necessary scaffolding is only available in Flink 1.16 and above.

ypark2103:
@reta I heard java 8 was deprecated starting Flink 1.15. Does it mean I need to upgrade to java 11 to use this connector? Or can I still use java 8 for Flink 1.16?

reta:
> @reta I heard java 8 was deprecated starting Flink 1.15. Does it mean I need to upgrade to java 11 to use this connector? Or can I still use java 8 for Flink 1.16?

@ypark2103 the connector itself has baseline of JDK-8 so you don't need JDK-11 to use it as-is (using 1.3.x OpenSearch client), but if you plan to upgrade to JDK-11, it is even better.

PR #3, opened 2022-12-08
[FLINK-25756] [connectors/opensearch] Dedicated Opensearch connectors (documentation)

reta:
Added documentation


reta:
@zentol here is the documentation part, I could bundle it into https://github.com/apache/flink-connector-opensearch/pull/1 if it does not complicate the review, thanks

docs/content/docs/connectors/datastream/opensearch.md
```diff
@@ -0,0 +1,328 @@
+---
+title: Opensearch
+weight: 5
+type: docs
+aliases:
+  - /dev/connectors/opensearch.html
+  - /apis/streaming/connectors/opensearch.html
```
zentol, line 7: you won't need these since the page didn't exist in the old structures :)

docs/content/docs/connectors/table/opensearch.md
```diff
@@ -0,0 +1,286 @@
...
+    <tr>
+      <td><h5>password</h5></td>
+      <td>optional</td>
+      <td>yes</td>
+      <td style="word-wrap: break-word;">(none)</td>
+      <td>String</td>
+      <td>Password used to connect to Opensearch instance. If <code>username</code> is configured, this option must be configured with non-empty string as well.</td>
+    </tr>
+    <tr>
+      <td><h5>failure-handler</h5></td>
```
zentol, line 140: This option doesn't exist in the table connector. you opted for using the non-deprecated connector from the table api which doesn't support the failure handler-

docs/content/docs/connectors/table/opensearch.md
```diff
@@ -0,0 +1,286 @@
...
+  pv BIGINT,
+  PRIMARY KEY (user_id) NOT ENFORCED
+) WITH (
+  'connector' = 'opensearch',
+  'hosts' = 'http://localhost:9200',
+  'index' = 'users'
+);
+```
+
+Connector Options
```
zentol, line 65: missing connection.timeout, socket.timeout, and possibly others.
This must be in sync with `OpensearchConnectorOptions`.

docs/content/docs/connectors/datastream/opensearch.md
```diff
@@ -0,0 +1,328 @@
...
+    .setHosts(new HttpHost("127.0.0.1", 9200, "http"))
+    .setEmitter((element: String, context: SinkWriter.Context, indexer: RequestIndexer) => 
+    indexer.add(createIndexRequest(element)))
+    // This enables an exponential backoff retry mechanism, with a maximum of 5 retries and an initial delay of 1000 milliseconds
+    .setBulkFlushBackoffStrategy(FlushBackoffType.EXPONENTIAL, 5, 1000)
+    .build())
+```
+
+{{< /tab >}}
+{{< tab "Python" >}}
```
zentol, line 273: The python API doesn't support this connector. (Needs a compatibility layer that currently has be added to Flink)

docs/content/docs/connectors/datastream/opensearch.md
```diff
@@ -0,0 +1,328 @@
...
+so by waiting for all pending action requests in the `BulkProcessor` at the
+time of checkpoints. This effectively assures that all requests before the
+checkpoint was triggered have been successfully acknowledged by Opensearch, before
+proceeding to process more records sent to the sink.
+
+More details on checkpoints and fault tolerance are in the [fault tolerance docs]({{< ref "docs/learn-flink/fault_tolerance" >}}).
+
+To use fault tolerant Opensearch Sinks, checkpointing of the topology needs to be enabled at the execution environment:
+
+{{< tabs "d00d1e93-4844-40d7-b0ec-9ec37e73145e" >}}
```
zentol, line 199: This should use a different id. Can be anything really; just make it different than the ES one.

docs/content/docs/connectors/datastream/opensearch.md
```diff
@@ -0,0 +1,328 @@
...
+        <td>{{< connector_artifact flink-connector-opensearch 1.0.0 >}}</td>
+    </tr>
+    <tr>
+        <td>2.x</td>
+        <td>{{< connector_artifact flink-connector-opensearch 1.0.0 >}}</td>
+    </tr>
+  </tbody>
+</table>
+
+{{< py_download_link "opensearch" >}}
```
zentol, line 53: ditto re python (see below)

zentol (changes requested):
In the original PR you mentioned this:
```
any mentions and uses of mapping types have been removed: it is deprecated feature, scheduled for removal (the indices with mapping types cannot be created or migrated to Opensearch 1.x and beyond)
any mentions and uses have been removed: it is deprecated feature, scheduled for removal (only HighLevelRestClient is used)
```
Was this documented anywhere?

reta:
> In the original PR you mentioned this:
> 
> ```
> any mentions and uses of mapping types have been removed: it is deprecated feature, scheduled for removal (the indices with mapping types cannot be created or migrated to Opensearch 1.x and beyond)
> any mentions and uses have been removed: it is deprecated feature, scheduled for removal (only HighLevelRestClient is used)
> ```
> 
> Was this documented anywhere?

Correct, it is not mentioned because the mapping types are not exposed anywhere anymore (so there is no way to use them). The Elasticsearch has explicit configuration setting for it https://github.com/apache/flink-connector-elasticsearch/blob/main/docs/content/docs/connectors/table/elasticsearch.md#document-type, it is not present for Opensearch).

docs/content/docs/connectors/datastream/opensearch.md
reta: Gone, thanks

docs/content/docs/connectors/table/opensearch.md
reta: Missed that, thank you

docs/content/docs/connectors/datastream/opensearch.md
reta: Removed

docs/content/docs/connectors/table/opensearch.md
reta: Fixed, I will submit the pull request to Elasticsearch connector to add those + delivery guarantee

flink-connector-opensearch/src/main/java/org/apache/flink/connector/opensearch/table/OpensearchConnectorOptions.java
```diff
@@ -143,7 +143,7 @@ public class OpensearchConnectorOptions {
     public static final ConfigOption<DeliveryGuarantee> DELIVERY_GUARANTEE_OPTION =
             ConfigOptions.key("sink.delivery-guarantee")
                     .enumType(DeliveryGuarantee.class)
-                    .defaultValue(DeliveryGuarantee.NONE)
+                    .defaultValue(DeliveryGuarantee.AT_LEAST_ONCE)
```
reta, line 146: Updating to match the right default

docs/content/docs/connectors/table/opensearch.md
reta: https://github.com/apache/flink-connector-elasticsearch/pull/49

reta: https://github.com/apache/flink-connector-elasticsearch/pull/49

PR #4, opened 2022-12-29
Update Opensearch dependency to version 2.3

amirhalatzi:
Hi!

During a project we're working on we decided to use Flink over Opensearch 2.3, which seems to have a breaking change of removing the `type` property from responses.

This PR updates the dependency and the relevant tests.

I'd be happy to do any necessary changes to make sure this release is differentiated from the one meant for Opensearch 1.3.

Thanks!

### DIFF
--- a/flink-connector-opensearch/pom.xml
+++ b/flink-connector-opensearch/pom.xml
@@ -36,7 +36,7 @@ under the License.
 
 	<!-- Allow users to pass custom connector versions -->
 	<properties>
-		<opensearch.version>1.3.0</opensearch.version>
+		<opensearch.version>2.3.0</opensearch.version>
 	</properties>
 
 	<dependencies>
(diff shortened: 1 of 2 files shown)


reta:
@amirhalatzi the OpenSearch 2.x has JDK-11 baseline (the connector works with 2.x as well). Please correct me if I am wrong, Apache Flink still supports JDK 8 (although it is deprecated).

reta:
> @reta you're right. Java 8 is still [supported](https://issues.apache.org/jira/browse/FLINK-25244). As for the current connector working with 2.x - you can see in the PR that the API has changed. Any ideas on how to support both versions without upgrading to Java 11?

@amirhalatzi this is only test code which uses `type`, sadly the Bulk API in 1.x requires it (at least for responses), it might be possible to . The connector's code itself does not use any deprecated APIs though.

PR #5, opened 2022-12-30
[FLINK-30488] OpenSearch implementation of Async Sink

reta:
OpenSearch implementation of Async Sink (https://cwiki.apache.org/confluence/display/FLINK/FLIP-171%3A+Async+Sink), a few TODO items:

- [ ] More test cases
- [ ]  More sensitive defaults


flink-connector-opensearch/src/test/java/org/apache/flink/connector/opensearch/sink/OpensearchSinkITCase.java
```diff
@@ -146,29 +146,30 @@ private void runTest(
                         .setAllowInsecure(true)
                         .build();
 
-        final StreamExecutionEnvironment env = new LocalStreamEnvironment();
-        env.enableCheckpointing(100L);
-        if (!allowRestarts) {
-            env.setRestartStrategy(RestartStrategies.noRestart());
-        }
-        DataStream<Long> stream = env.fromSequence(1, 5);
+        try (final StreamExecutionEnvironment env = new LocalStreamEnvironment()) {
```
reta, line 149: Not related by `StreamExecutionEnvironment` is `AutoCloseable`, slightly changing the test case

reta:
@zentol @MartijnVisser would appreciate if you guys have time for the review, adding AsyncSink support for OpenSearch, discussed initially [1]

[1] https://github.com/apache/flink/pull/18541#issuecomment-1026931087

reta:
@zentol @MartijnVisser doing my one per month ping diligence :-), please

MartijnVisser:
> @zentol @MartijnVisser doing my one per month ping diligence :-), please

I'm currently a bit over capacity. Don't know if the same applies for @zentol tbh

@dannycranmer Could you potentially help out? You also have the experience with the Async API, or perhaps @hlteoh37 ?

hlteoh37:
Sure, I can take a look

flink-connector-opensearch-e2e-tests/src/main/java/org/apache/flink/streaming/tests/OpensearchAsyncSinkExample.java
```diff
@@ -0,0 +1,81 @@
...
+                                            Long value, Collector<Tuple2<String, String>> out) {
+                                        final String key = String.valueOf(value);
+                                        final String message = "message #" + value;
+                                        out.collect(Tuple2.of(key, message + "update #1"));
+                                        out.collect(Tuple2.of(key, message + "update #2"));
+                                    }
+                                });
+
+        List<HttpHost> httpHosts = new ArrayList<>();
+        httpHosts.add(new HttpHost("127.0.0.1", 9200, "http"));
```
hlteoh37, line 66: not sure what this is used for, shall we remove it?

reta: That's just an example of using AsyncSync, we used to have them `https://github.com/apache/flink-connector-opensearch/tree/main/flink-connector-opensearch-e2e-tests/src/main/java/org/apache/flink/streaming/tests`

flink-connector-opensearch/src/main/java/org/apache/flink/connector/opensearch/sink/DocSerdeRequest.java
```diff
@@ -0,0 +1,107 @@
...
+        if (type == 0) {
+            docWriteRequest = new IndexRequest(in);
+        } else if (type == 1) {
+            docWriteRequest = new DeleteRequest(in);
+        } else if (type == 2) {
+            docWriteRequest = new UpdateRequest(in);
+        } else {
+            throw new IllegalStateException("Invalid request type [" + type + " ]");
+        }
+        return docWriteRequest;
```
hlteoh37, line 87: These methods are untested. Should we add unit tests for them?

flink-connector-opensearch-e2e-tests/src/main/java/org/apache/flink/streaming/tests/OpensearchAsyncSinkExample.java
```diff
@@ -0,0 +1,81 @@
...
+
+        OpensearchAsyncSinkBuilder<Tuple2<String, String>> osSinkBuilder =
+                OpensearchAsyncSink.<Tuple2<String, String>>builder()
+                        .setHosts(new HttpHost("localhost:9200"))
+                        .setElementConverter(
+                                (element, context) ->
+                                        new IndexRequest("my-index")
+                                                .id(element.f0.toString())
+                                                .source(element.f1));
```
hlteoh37, line 75: Hmm.. Since we have to implement a `DocSerdeRequest`, should we consider exposing this in the interface instead of `OpenSearch` classes? This might be helpful in the event OpenSearch's interface changes.

hlteoh37 (changes requested):
Added some comments!

flink-connector-opensearch-e2e-tests/src/main/java/org/apache/flink/streaming/tests/OpensearchAsyncSinkExample.java
```diff
@@ -0,0 +1,81 @@
...
+                                        out.collect(Tuple2.of(key, message + "update #2"));
+                                    }
+                                });
+
+        List<HttpHost> httpHosts = new ArrayList<>();
+        httpHosts.add(new HttpHost("127.0.0.1", 9200, "http"));
+
+        OpensearchAsyncSinkBuilder<Tuple2<String, String>> osSinkBuilder =
+                OpensearchAsyncSink.<Tuple2<String, String>>builder()
+                        .setHosts(new HttpHost("localhost:9200"))
```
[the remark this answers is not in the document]
reta: Oh I see, there are 2 places where same host is used, should be 1, I will fix that, thank you

flink-connector-opensearch/src/main/java/org/apache/flink/connector/opensearch/sink/OpensearchAsyncSink.java
```diff
@@ -0,0 +1,150 @@
...
+    private static final long serialVersionUID = 1L;
+
+    private final List<HttpHost> hosts;
+    private final NetworkClientConfig networkClientConfig;
+
+    /**
+     * Constructor creating an Opensearch async sink.
+     *
+     * @param maxBatchSize the maximum size of a batch of entries that may be sent
+     * @param maxInFlightRequests he maximum number of in flight requests that may exist, if any
```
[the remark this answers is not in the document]
reta: :+1:

flink-connector-opensearch/src/main/java/org/apache/flink/connector/opensearch/sink/OpensearchAsyncSink.java
```diff
@@ -0,0 +1,150 @@
...
+     *     more in flight requests need to be initiated once the maximum has been reached, then it
+     *     will be blocked until some have completed
+     * @param maxBufferedRequests the maximum number of elements held in the buffer, requests to add
+     *     elements will be blocked while the number of elements in the buffer is at the maximum
+     * @param maxBatchSizeInBytes the maximum size of a batch of entries that may be sent to KDS
```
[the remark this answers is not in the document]
reta: :+1:

flink-connector-opensearch/src/main/java/org/apache/flink/connector/opensearch/sink/OpensearchAsyncSinkBuilder.java
```diff
@@ -0,0 +1,220 @@
...
+
+    /**
+     * Sets the hosts where the Opensearch cluster nodes are reachable.
+     *
+     * @param hosts http addresses describing the node locations
+     * @return this builder
+     */
+    public OpensearchAsyncSinkBuilder<InputT> setHosts(HttpHost... hosts) {
+        checkNotNull(hosts);
+        checkState(hosts.length > 0, "Hosts cannot be empty.");
```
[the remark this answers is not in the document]
reta: I thing checking twice is acceptable here: we should fail as early as possible, allowing to construct a builder with possibly illegal arguments and carrying it over could potentially raise an exception down the stack, when `build` method is called. By validating early, we are preventing that.

flink-connector-opensearch/src/main/java/org/apache/flink/connector/opensearch/sink/OpensearchAsyncSinkBuilder.java
```diff
@@ -0,0 +1,220 @@
...
+    }
+
+    /**
+     * Sets the timeout for requesting the connection of the Opensearch cluster from the connection
+     * manager.
+     *
+     * @param timeout for the connection request
+     * @return this builder
+     */
+    public OpensearchAsyncSinkBuilder<InputT> setConnectionRequestTimeout(int timeout) {
```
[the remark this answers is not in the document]
reta: We have these properties documented and use the same settings across all sinks and configurations, but I could surely add the overloaded method for convenience.

flink-connector-opensearch/src/main/java/org/apache/flink/connector/opensearch/sink/OpensearchAsyncSinkBuilder.java
```diff
@@ -0,0 +1,220 @@
...
+    }
+
+    /**
+     * Sets the timeout for waiting for data or, put differently, a maximum period inactivity
+     * between two consecutive data packets.
+     *
+     * @param timeout for the socket
+     * @return this builder
+     */
+    public OpensearchAsyncSinkBuilder<InputT> setSocketTimeout(int timeout) {
```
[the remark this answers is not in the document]
reta: Added another method for convience

flink-connector-opensearch/src/main/java/org/apache/flink/connector/opensearch/sink/OpensearchAsyncSinkBuilder.java
```diff
@@ -0,0 +1,220 @@
...
+    }
+
+    @Override
+    public OpensearchAsyncSink<InputT> build() {
+        checkArgument(!hosts.isEmpty(), "Hosts cannot be empty.");
+
+        return new OpensearchAsyncSink<InputT>(
+                nonNullOrDefault(
+                        getMaxBatchSize(),
+                        1000), /* OpensearchConnectorOptions.BULK_FLUSH_MAX_ACTIONS_OPTION */
```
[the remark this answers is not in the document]
reta: Creating `Optiona`l (new objects) just for convenience of checkiing for availability should be avoided.

flink-connector-opensearch/src/main/java/org/apache/flink/connector/opensearch/sink/OpensearchAsyncSinkBuilder.java
```diff
@@ -0,0 +1,220 @@
...
+                nonNullOrDefault(
+                        getMaxInFlightRequests(), 1), /* BulkProcessor::concurrentRequests */
+                nonNullOrDefault(getMaxBufferedRequests(), 10000),
+                nonNullOrDefault(
+                        getMaxBatchSizeInBytes(),
+                        2 * 1024
+                                * 1024), /* OpensearchConnectorOptions.BULK_FLUSH_MAX_SIZE_OPTION */
+                nonNullOrDefault(
```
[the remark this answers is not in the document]
reta: This is specific to AWS OpenSearch managed service, this is not applicable to OpenSearch in general.

flink-connector-opensearch/src/main/java/org/apache/flink/connector/opensearch/sink/OpensearchAsyncSink.java
```diff
@@ -0,0 +1,150 @@
...
+     * @param networkClientConfig describing properties of the network connection used to connect to
+     *     the Opensearch cluster
+     */
+    OpensearchAsyncSink(
+            int maxBatchSize,
+            int maxInFlightRequests,
+            int maxBufferedRequests,
+            long maxBatchSizeInBytes,
+            long maxTimeInBufferMS,
+            long maxRecordSizeInBytes,
```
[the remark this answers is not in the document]
reta: This is specific to AWS OpenSearch managed service, this is not applicable to OpenSearch in general.

flink-connector-opensearch/src/main/java/org/apache/flink/connector/opensearch/sink/OpensearchAsyncWriter.java
```diff
@@ -0,0 +1,274 @@
...
+                            requestConfigBuilder.setConnectTimeout(
+                                    networkClientConfig.getConnectionTimeout());
+                        }
+                        if (networkClientConfig.getSocketTimeout() != null) {
+                            requestConfigBuilder.setSocketTimeout(
+                                    networkClientConfig.getSocketTimeout());
+                        }
+                        return requestConfigBuilder;
+                    });
+        }
```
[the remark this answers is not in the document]
reta: The presence of first `if` helps to elimination the need to create the `RequestConfigCallback` instance at all if there is nothing to configure.

flink-connector-opensearch-e2e-tests/src/main/java/org/apache/flink/streaming/tests/OpensearchAsyncSinkExample.java
reta: The `DocSerdeRequest` is sadly a necessary leaking abstraction (AsyncSink requires `Serializable`), we should export in the places when it is inevitable but in general we should only operate over OpenSearch APIs.

flink-connector-opensearch/src/main/java/org/apache/flink/connector/opensearch/sink/DocSerdeRequest.java
reta: They are tested in scope if integration test, OpensearchAsyncSinkITCase, both reading and writing side.

flink-connector-opensearch/src/main/java/org/apache/flink/connector/opensearch/sink/OpensearchAsyncWriter.java
```diff
@@ -0,0 +1,274 @@
...
+    private void handlePartiallyFailedBulkRequests(
+            BulkResponse response,
+            List<DocSerdeRequest<?>> requestEntries,
+            Consumer<List<DocSerdeRequest<?>>> requestResult) {
+
+        final List<DocSerdeRequest<?>> failedRequestEntries = new ArrayList<>();
+        final BulkItemResponse[] items = response.getItems();
+
+        for (int i = 0; i < items.length; i++) {
+            if (items[i].getFailure() != null) {
```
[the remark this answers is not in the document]
reta: I don't think so, there could be massive amount of items in bulk request, logging 10k failures (failure is reported per item), would probably flood the logs

flink-connector-opensearch/src/main/java/org/apache/flink/connector/opensearch/sink/OpensearchAsyncWriter.java
```diff
@@ -0,0 +1,274 @@
...
+            try {
+                client.close();
+            } catch (final IOException ex) {
+                LOG.warn("Error while closing RestHighLevelClient instance", ex);
+            }
+        }
+    }
+
+    private void handleFullyFailedBulkRequest(
+            Throwable err,
```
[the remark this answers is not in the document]
reta: Certainly makes sense!

.github/workflows/push_pr.yml
```diff
@@ -25,6 +25,6 @@ jobs:
   compile_and_test:
     uses: apache/flink-connector-shared-utils/.github/workflows/ci.yml@ci_utils
     with:
-      flink_version: 1.16.0
-      flink_url: https://dist.apache.org/repos/dist/release/flink/flink-1.16.0/flink-1.16.0-bin-scala_2.12.tgz
+      flink_version: 1.16.1
+      flink_url: https://dist.apache.org/repos/dist/release/flink/flink-1.16.1/flink-1.16.1-bin-scala_2.12.tgz
```
reta, line 29: `https://dist.apache.org/repos/dist/release/flink/flink-1.16.0/flink-1.16.0-bin-scala_2.12.tgz` is not available anymore :-|

reta:
Thanks a lot for the review @hlteoh37 , I believe I addressed or/and answered all your comments, please let me know if I missed something

.github/workflows/push_pr.yml
MartijnVisser: Have a look at https://github.com/apache/flink-connector-jdbc/pull/21 with how we've solved it :)

reta: Thanks @MartijnVisser , still needs version change, right? 1.16.0 -> 1.16.1?

MartijnVisser: > Thanks @MartijnVisser , still needs version change, right? 1.16.0 -> 1.16.1?

It depends if you want to support 1.16.1 (I think you would). But if you would like to remain on 1.16.0, at least the link isn't dead anymore :)

reta: Gotcha, thank you! (yes, I think we should support 1.16.1)

.github/workflows/push_pr.yml
```diff
@@ -25,4 +25,4 @@ jobs:
   compile_and_test:
     uses: apache/flink-connector-shared-utils/.github/workflows/ci.yml@ci_utils
     with:
-      flink_version: 1.16.0
+      flink_version: 1.16.1
```
dannycranmer, line 28: We should split this out to a separate PR. We can merge this independently

flink-connector-opensearch-e2e-tests/src/main/java/org/apache/flink/streaming/tests/OpensearchAsyncSinkExample.java
dannycranmer: This is a shame indeed, because Async Sink does not actually need Serializable. https://issues.apache.org/jira/browse/FLINK-27537

flink-connector-opensearch/src/main/java/org/apache/flink/connector/opensearch/sink/DocSerdeRequest.java
```diff
@@ -0,0 +1,107 @@
...
+import java.io.Serializable;
+
+/**
+ * Wrapper class around {@link DocWriteRequest} since it does not implement {@link Serializable},
+ * required by AsyncSink scaffolding.
+ *
+ * @param <T> type of the write request
+ */
+@PublicEvolving
+public class DocSerdeRequest<T> implements Serializable {
```
dannycranmer, line 45: Ii think the class level generics are redundant here. We are using `<?>` throughout. Consider changing `private final DocWriteRequest<T> request;` to `private final DocWriteRequest<?> request;` and removing class generics. This makes the Sink interface a bit messy `extends AsyncSinkBase<InputT, DocSerdeRequest<?>>`

flink-connector-opensearch/src/main/java/org/apache/flink/connector/opensearch/sink/DocSerdeRequest.java
dannycranmer: +1 for a unit test. Unless there is a good reason not to, unit tests give quicker feedback.

flink-connector-opensearch/src/main/java/org/apache/flink/connector/opensearch/sink/OpensearchAsyncSinkBuilder.java
```diff
@@ -0,0 +1,246 @@
...
+
+    @Override
+    public OpensearchAsyncSink<InputT> build() {
+        return new OpensearchAsyncSink<InputT>(
+                nonNullOrDefault(
+                        getMaxBatchSize(),
+                        1000), /* OpensearchConnectorOptions.BULK_FLUSH_MAX_ACTIONS_OPTION */
+                nonNullOrDefault(
+                        getMaxInFlightRequests(), 1), /* BulkProcessor::concurrentRequests */
+                nonNullOrDefault(getMaxBufferedRequests(), 10000),
```
dannycranmer, line 214: Can we also promote the other magic numbers to constants? `10000` and `2 * 1024 * 1024`

flink-connector-opensearch/src/main/java/org/apache/flink/connector/opensearch/sink/OpensearchAsyncWriter.java
```diff
@@ -0,0 +1,308 @@
...
+                                            .build());
+                        } catch (final NoSuchAlgorithmException
+                                | KeyStoreException
+                                | KeyManagementException ex) {
+                            throw new IllegalStateException(
+                                    "Unable to create custom SSL context", ex);
+                        }
+                    }
+
+                    return httpClientBuilder;
```
dannycranmer, line 284: nit: Should we move this out to a separate class?

dannycranmer (changes requested):

.github/workflows/push_pr.yml
reta: @dannycranmer https://github.com/apache/flink-connector-opensearch/pull/7

flink-connector-opensearch/src/main/java/org/apache/flink/connector/opensearch/sink/DocSerdeRequest.java
reta: This is my bad, the `T` must be constrained, I will fix it

flink-connector-opensearch/src/main/java/org/apache/flink/connector/opensearch/sink/OpensearchAsyncWriter.java
reta: why? that's the only place is needed actually, seems like sealing it in place is acceptable

reta:
Thanks @dannycranmer , I think I went through all your comments, thanks a lot, really appreciate it.

.github/workflows/push_pr.yml
reta: (addressed in https://github.com/apache/flink-connector-opensearch/pull/7)

flink-connector-opensearch/src/main/java/org/apache/flink/connector/opensearch/sink/OpensearchAsyncWriter.java
dannycranmer: In my opinion it breaches the single responsibility philosophy. The writer is responsible for writing and knowing how to construct the client. I am less concerned how many times it is used. However, I marked as _nit_ since I am not marking this as a blocker.

reta: Thanks @dannycranmer , I think this is a good idea (more over, I was not correct, there was another place with the similar instantiation logic present), extracted the utility class

flink-connector-opensearch/src/test/java/org/apache/flink/connector/opensearch/sink/DocSerdeRequestTest.java
```diff
@@ -0,0 +1,92 @@
...
+                                        .build())
+                        .isEqualTo(serialized.getRequest());
+            }
+        }
+    }
+
+    @Test
+    @SuppressWarnings("unchecked")
+    void unsupportedRequestType() throws IOException {
+        final DocSerdeRequest serialized = DocSerdeRequest.from(mock(DocWriteRequest.class));
```
dannycranmer, line 70: Mockito is [banned](https://flink.apache.org/how-to-contribute/code-style-and-quality-common/#avoid-mockito---use-reusable-test-implementations). Since you only have one usage here can we remove it?

dannycranmer:
@reta The PR looks good to me minus the Mockito comment. However I have questions over the approach here. We are adding a new sink alongside the existing sink, we will have `OpensearchSink` and `OpensearchAsyncSink`. How do the users know which one to pick? Why not replace the existing sink with the new implementation? The [Jira](https://issues.apache.org/jira/browse/FLINK-31068) mentions docs, however there is no update here. Will you create a followup PR for that?

If this has already been discussed on mailing lists I missed that, please give me a link :D

reta:
Thanks a lot for review @dannycranmer 

> @reta The PR looks good to me minus the Mockito comment. However I have questions over the approach here. We are adding a new sink alongside the existing sink, we will have `OpensearchSink` and `OpensearchAsyncSink`. How do the users know which one to pick? Why not replace the existing sink with the new implementation? 

This is a indeed a good question, I think the main difference between those are within internal APIs the implementation is based upon:
 - `OpensearchAsyncSink` uses `RestHighLevelClient::bulkdAsync` directly to dispatch the bulk requests
 - `OpensearchSink` uses `BulkProcessor` and offers more flexibility with respect to failure handling and backoff policies (no straight equivalent in `RestHighLevelClient`)

I have covered this part in the docs, thank you.

> The [Jira](https://issues.apache.org/jira/browse/FLINK-31068) mentions docs, however there is no update here. Will you create a followup PR for that?
> 

Updated the documentation, thank you

> If this has already been discussed on mailing lists I missed that, please give me a link :D

You mean the `AsyncSync` implementation for OpenSearch? No, it was not discussed on mailing list but was mentioned on the initial pull request https://github.com/apache/flink/pull/18541#issuecomment-1026931087

reta:
@dannycranmer would appreciate if you could take a look, thank you

dannycranmer:
@reta I am reluctant to introduce a new Sink API based on the internal implementation unless there is a really good/semantic reason. I would prefer to encapsulate the internals via a single Flink layer that can support either `RestHighLevelClient`/`BulkProcessor` based on configuration. How will this look for SQL? We usually use a simple identifier like "opensearch", I fear that "opensearch-async" adds no semantic value to the user.

We should keep the Sink API as simple as possible with sensible defaults, and allow advanced users to configure as they wish. For instance, a user should not need decide to use `OpensearchSink` vs `OpensearchAsyncSink`, they should just use `OpensearchSink` and configure as needed.

There could be reasons to have 2x Sinks if they support fundamentally different features/APIs but I would expect the naming to reflect this, for example `OpensearchRestHighLevelClientSink`/`OpensearchBulkProcessorSink`. 

Apologies for raising these concerns late in the process but I cannot see this has been considered before. @MartijnVisser what are your thoughts?

reta:
> I am reluctant to introduce a new Sink API based on the internal implementation unless there is a really good/semantic reason.

Thanks @dannycranmer , I understand your concerns. I will move this pull request to draft (for now) so we could get to it at some point in the future, when migrating off the `RestHighLevelClient` to `opensearch-java`, thanks again for review and your thoughts.

PR #8, opened 2023-02-12
Add FailureHandler.

lilyevsky:
Here are my changes to add the FailureHandler.
I also did minor fixes in the tests, related to the changes in the Opensearch 2.5.0 (otherwise they did not compile).
In the OpensearchWriterITCase.java:269 I set the failureHandler to null, so the default behavior will be tested. I hope it is OK, because I did test the actual handler in the real runtime environment. However, we can add the actual handler test if necessary.

Also, let's discuss whether I put the FailureHandler interface in the proper package. This interface is pretty generic by its nature, so maybe you want to have it in some other package.

### DIFF
--- a/flink-connector-opensearch/src/main/java/org/apache/flink/connector/opensearch/sink/FailureHandler.java
+++ b/flink-connector-opensearch/src/main/java/org/apache/flink/connector/opensearch/sink/FailureHandler.java
@@ -1,0 +1,26 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one
+ * or more contributor license agreements.  See the NOTICE file
+ * distributed with this work for additional information
+ * regarding copyright ownership.  The ASF licenses this file
+ * to you under the Apache License, Version 2.0 (the
+ * "License"); you may not use this file except in compliance
+ * with the License.  You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package org.apache.flink.connector.opensearch.sink;
+
+import java.io.Serializable;
+
+public interface FailureHandler extends Serializable {
+    void onFailure(Throwable failure);
+}
+
--- a/flink-connector-opensearch/src/main/java/org/apache/flink/connector/opensearch/sink/FailureHandler.java
+++ b/flink-connector-opensearch/src/main/java/org/apache/flink/connector/opensearch/sink/FailureHandler.java
@@ -20,6 +20,9 @@ package org.apache.flink.connector.opensearch.sink;
 
 import java.io.Serializable;
 
+/**
+ * Handler to process failures.
+ */
 public interface FailureHandler extends Serializable {
     void onFailure(Throwable failure);
 }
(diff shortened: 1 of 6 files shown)


flink-connector-opensearch/src/main/java/org/apache/flink/connector/opensearch/sink/FailureHandler.java
```diff
@@ -0,0 +1,26 @@
...
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package org.apache.flink.connector.opensearch.sink;
+
+import java.io.Serializable;
+
+/** Handler to process failures. */
+public interface FailureHandler extends Serializable {
```
reta, line 24: Could you please add `@FunctionalInterface` to this interface (for usability perspective).

flink-connector-opensearch/src/main/java/org/apache/flink/connector/opensearch/sink/OpensearchSink.java
```diff
@@ -59,19 +59,22 @@
     private final BulkProcessorConfig buildBulkProcessorConfig;
     private final NetworkClientConfig networkClientConfig;
     private final DeliveryGuarantee deliveryGuarantee;
+    private FailureHandler failureHandler;
```
reta, line 62: `private final` please

flink-connector-opensearch/src/main/java/org/apache/flink/connector/opensearch/sink/OpensearchSinkBuilder.java
```diff
@@ -285,6 +286,18 @@ public OpensearchSinkBuilder<IN> setAllowInsecure(boolean allowInsecure) {
...
 
+    /**
+     * Allows to set custom failure handler. If not set, or set to null, then the default behavior
+     * is preserved: throwing a runtime exception upon receiving a failure.
+     *
+     * @param failureHandler
+     * @return
+     */
+    public OpensearchSinkBuilder<IN> setFailureHandler(FailureHandler failureHandler) {
+        this.failureHandler = failureHandler;
```
reta, line 297: Please `checkNotNull(failureHandler);`

flink-connector-opensearch/src/main/java/org/apache/flink/connector/opensearch/sink/OpensearchWriter.java
```diff
@@ -343,7 +346,11 @@ private void extractFailures(BulkRequest request, BulkResponse response) {
         if (chainedFailures == null) {
             return;
         }
-        throw new FlinkRuntimeException(chainedFailures);
+        if (failureHandler == null) {
```
reta, line 349: Dealing with`null` is better to be avoided: we could use the default `failureHandler`:

```java
private static final FailureHandler DEFAULT_FAILURE_HANDLER = ex -> throw new FlinkRuntimeException(ex);
```

and never deal with nullability / `if`s

flink-connector-opensearch/src/test/java/org/apache/flink/connector/opensearch/sink/OpensearchWriterITCase.java
```diff
@@ -266,7 +266,8 @@ private OpensearchWriter<Tuple2<Integer, String>> createWriter(
                         null,
                         true),
                 metricGroup,
-                new TestMailbox());
+                new TestMailbox(),
+                null);
```
reta, line 270: `DEFAULT_FAILURE_HANDLER` would help here to eliminate `null`

reta:
@lilyevsky could you please retarget the pull request to `main` ? thank you.

PR #11, opened 2023-02-15
[FLINK-30998] Add optional exception handler to flink-connector-opensearch

lilyevsky:


reta:
@lilyevsky the relevant changes looks good, but your branch is out of sync with https://github.com/apache/flink-connector-opensearch, you need to sync + rebase

lilyevsky:
@reta I guess I did something wrong with my repo; I made a minor change in the comments, so I wanted to squish the commits and force push. I did rebase, I thought, against your commit that I saw in the history. I thought then PR would only show my one commit.
It now shows some other commits that, I believe, were already merged into apache:main.
Please let me know if this is a problem. Worst case, I can do it all over again clean.

lilyevsky:
I think I can do it in github: https://docs.github.com/en/pull-requests/collaborating-with-pull-requests/working-with-forks/syncing-a-fork

flink-connector-opensearch/src/main/java/org/apache/flink/connector/opensearch/sink/OpensearchSink.java
```diff
@@ -59,19 +59,22 @@
...
-            NetworkClientConfig networkClientConfig) {
+            NetworkClientConfig networkClientConfig,
+            FailureHandler failureHandler) {
         this.hosts = checkNotNull(hosts);
         checkArgument(!hosts.isEmpty(), "Hosts cannot be empty.");
         this.emitter = checkNotNull(emitter);
         this.deliveryGuarantee = checkNotNull(deliveryGuarantee);
         this.buildBulkProcessorConfig = checkNotNull(buildBulkProcessorConfig);
         this.networkClientConfig = checkNotNull(networkClientConfig);
+        this.failureHandler = failureHandler;
```
reta, line 77:
```suggestion
        this.failureHandler = checkNotNull(failureHandler);
```

flink-connector-opensearch/src/main/java/org/apache/flink/connector/opensearch/sink/OpensearchSinkBuilder.java
```diff
@@ -72,6 +72,7 @@
     private Integer connectionRequestTimeout;
     private Integer socketTimeout;
     private Boolean allowInsecure;
+    private FailureHandler failureHandler;
```
reta, line 75:
```suggestion
    private FailureHandler failureHandler = DEFAULT_FAILURE_HANDLER; 
```

flink-connector-opensearch/src/main/java/org/apache/flink/connector/opensearch/sink/OpensearchWriter.java
```diff
@@ -343,7 +348,11 @@ private void extractFailures(BulkRequest request, BulkResponse response) {
         if (chainedFailures == null) {
             return;
         }
-        throw new FlinkRuntimeException(chainedFailures);
+        if (failureHandler == null) {
```
reta, line 351: This `null` check is not needed anymore (the builder would supply the default handler), just `failureHandler.onFailure(chainedFailures);`

reta:
Thanks @lilyevsky just a few minor comments, I think the only part that is missed are tests, and probably you may need to run `mvn spotless:apply` to apply the formatting, thank you!

lilyevsky:
@reta Did all the changes you suggested, also ran the mvn spotless:apply .
Btw., my IntelliJ showed a list of warnings in places not relevant to my changes. Even one error (which is not really an error because it is in the comments), in OpensearchWriter.java:96.
Do you have any suggestions for the tests? As I mentioned before, I actually tested this in real environment; on the other hand, I understand that unit tests are very important to have.

reta:
> @reta Did all the changes you suggested, also ran the mvn spotless:apply . 

👍  thank you!

>  Even one error (which is not really an error because it is in the comments), in OpensearchWriter.java:96.

Hm .. could you share the error please?

> Do you have any suggestions for the tests? 

Yes, sure, I think [OpensearchWriterITCase](https://github.com/apache/flink-connector-opensearch/blob/main/flink-connector-opensearch/src/test/java/org/apache/flink/connector/opensearch/sink/OpensearchWriterITCase.java) could be a good candidate to add a test case with the simulated failure (fe we could pause / shutdown Opensearch container, just an idea). It may be a bit difficult, please let me know if help needed, thank you.

lilyevsky:
The error is about this line:

 * @param bulkProcessorBuilderFactory configuring the {@link BulkProcessor}'s builder

Apparently that parameter is no longer present. 
Also 10 warnings, like "Return value ... is never used" or "Exception ... is never thrown".
One warning is actually about failureHandler, missing description.

I actually can fix all of that if it is OK, then you can review.

Also will look into the tests tomorrow.

reta:
> I actually can fix all of that if it is OK, then you can review.

Thank you, cleanup would be much appreciated, I would suggest let's finish up with tests first so to have the scope finalized, and we could fix javadocs after. Thanks!

lilyevsky:
Sure. 
I will fix that bulkProcessorBuilderFactory  thing, everything else is less important at the moment. I got that list of warnings while committing. When I just run 'analyze', I get even more stuff.

lilyevsky:
@reta I fixed some warnings, pushed as a separate commit, so you can review.
Now looking at the tests. We need to somehow simulate the Opensearch server that can get the request and then send back the response. Is there currently such thing in the test suite? Is it what you do in flink-connector-opensearch-e2e-tests?
In any case, I think you would know better than I how to do it.

reta:
> @reta I fixed some warnings, pushed as a separate commit, so you can review. Now looking at the tests. We need to somehow simulate the Opensearch server that can get the request and then send back the response. 

I shared the hints with you here https://github.com/apache/flink-connector-opensearch/pull/11#issuecomment-1432308014

Here is sample test case:

```
    @Test
    void testWriteErrorOnUpdate() throws Exception {
        final String index = "test-bulk-flush-with-error";
        final int flushAfterNActions = 1;
        final BulkProcessorConfig bulkProcessorConfig =
                new BulkProcessorConfig(flushAfterNActions, -1, -1, FlushBackoffType.NONE, 0, 0);

        try (final OpensearchWriter<Tuple2<Integer, String>> writer =
                createWriter(index, true, bulkProcessorConfig)) {
            // Trigger an error by updating non-existing document
            writer.write(Tuple2.of(1, "u" + buildMessage(1)), null);
            context.assertThatIdsAreNotWritten(index, 1);
        }
    }
```

You need to attach the failure handler and assert afterwards it has been called.

lilyevsky:
@reta Trying to add the test. For some reason the tests from OpensearchWriterITCase class are not executed at all.
I run the command:

mvn clean test

in flink-connector-opensearch directory. I see that some tests from org.apache.flink.connector.opensearch.sink.OpensearchSinkBuilderTest are executed, but not from other 4 files in the same directory.
What could be wrong?

[maventest.txt](https://github.com/apache/flink-connector-opensearch/files/10760344/maventest.txt)

lilyevsky:
Thanks @reta . One more step. It looks like I need to have a "valid Docker environment". I install the docker things:

sudo apt-get install docker-ce docker-ce-cli containerd.io docker-buildx-plugin docker-compose-plugin

What else do I need to do?

lilyevsky:
@reta I installed docker, the daemon is running, the hello world test worked.
However, the 'mvn verify' still fails with "OpensearchSinkITCase » IllegalState Could not find a valid Docker environment...." ....
How does maven know where to look for Docker environment? Maybe I am some env variable?

lilyevsky:
@reta Added the test. I think now it is good.
I also tested the test itself, by simulating the "bad" handler, making the test fail, just to make sure it checks it properly.

flink-connector-opensearch/src/main/java/org/apache/flink/connector/opensearch/sink/OpensearchWriter.java
```diff
@@ -103,7 +107,8 @@
             BulkProcessorConfig bulkProcessorConfig,
             NetworkClientConfig networkClientConfig,
             SinkWriterMetricGroup metricGroup,
-            MailboxExecutor mailboxExecutor) {
+            MailboxExecutor mailboxExecutor,
```
reta, line 110: Please update javadocs with `failureHandler` arg

flink-connector-opensearch/src/test/java/org/apache/flink/connector/opensearch/sink/OpensearchWriterITCase.java
```diff
@@ -238,19 +240,68 @@ void testCurrentSendTime() throws Exception {
...
+
+        private synchronized void setFailed() {
+            failed = true;
+        }
+
+        public boolean isFailed() {
+            return failed;
+        }
+
+        @java.lang.Override
```
reta, line 254: Super nit, just `@Override`

flink-connector-opensearch/src/test/java/org/apache/flink/connector/opensearch/sink/OpensearchWriterITCase.java
```diff
@@ -238,19 +240,68 @@ void testCurrentSendTime() throws Exception {
         }
     }
 
+    private class TestHandler implements FailureHandler {
```
reta, line 243: Probably `private static` is better

flink-connector-opensearch/src/test/java/org/apache/flink/connector/opensearch/sink/OpensearchWriterITCase.java
```diff
@@ -238,19 +240,68 @@ void testCurrentSendTime() throws Exception {
...
+        final BulkProcessorConfig bulkProcessorConfig =
+                new BulkProcessorConfig(flushAfterNActions, -1, -1, FlushBackoffType.NONE, 0, 0);
+
+        final TestHandler testHandler = new TestHandler();
+        try (final OpensearchWriter<Tuple2<Integer, String>> writer =
+                createWriter(index, true, bulkProcessorConfig, testHandler)) {
+            // Trigger an error by updating non-existing document
+            writer.write(Tuple2.of(1, "u" + buildMessage(1)), null);
+            context.assertThatIdsAreNotWritten(index, 1);
+            assertThat(testHandler.isFailed()).isEqualTo(true);
```
reta, line 273: :+1:

flink-connector-opensearch/src/test/java/org/apache/flink/connector/opensearch/sink/OpensearchWriterITCase.java
```diff
@@ -174,7 +175,8 @@ void testIncrementByteOutMetric() throws Exception {
                 new BulkProcessorConfig(flushAfterNActions, -1, -1, FlushBackoffType.NONE, 0, 0);
 
         try (final OpensearchWriter<Tuple2<Integer, String>> writer =
-                createWriter(index, false, bulkProcessorConfig, metricGroup)) {
+                createWriter(
```
reta, line 178: You probably don't need this change since `createWriter` has an overloaded version without failure handler

flink-connector-opensearch/src/main/java/org/apache/flink/connector/opensearch/sink/FailureHandler.java
```diff
@@ -0,0 +1,30 @@
...
+ */
+
+package org.apache.flink.connector.opensearch.sink;
+
+import org.apache.flink.annotation.Public;
+
+import java.io.Serializable;
+
+/** Handler to process failures. */
+@Public
```
reta, line 26: `@Public` -> `@PublicEvolving` would be probably better (to leave some space to evolve the API)

reta:
@lilyevsky a few minor comments, LGTM otherwise!

flink-connector-opensearch/src/test/java/org/apache/flink/connector/opensearch/sink/OpensearchWriterITCase.java
lilyevsky: Agreed, it will be better not to touch that line. I did it because it needed to pass the metricGroup.
So I am fixing the situation by adding yet another constructor, with metricGroup but without failureHandler.

reta: No need for constructor, `createWriter` is a method in this class

lilyevsky:
@reta I committed the changes you suggested.

flink-connector-opensearch/src/main/java/org/apache/flink/connector/opensearch/sink/OpensearchWriter.java
```diff
@@ -122,10 +128,11 @@
         } catch (Exception e) {
             throw new FlinkRuntimeException("Failed to open the OpensearchEmitter", e);
         }
+        this.failureHandler = failureHandler;
     }
 
     @Override
-    public void write(IN element, Context context) throws IOException, InterruptedException {
+    public void write(IN element, Context context) throws InterruptedException {
```
reta, line 135: Please revert these changes, they violate the `SinkWriter` interface 

[1] https://github.com/apache/flink/blob/master/flink-core/src/main/java/org/apache/flink/api/connector/sink2/SinkWriter.java#L41

flink-connector-opensearch/src/main/java/org/apache/flink/connector/opensearch/sink/OpensearchSinkBuilder.java
```diff
@@ -148,6 +150,7 @@ public OpensearchSinkBuilder<IN> setBulkFlushMaxActions(int numMaxActions) {
      * @param maxSizeMb the maximum size of buffered actions, in mb.
      * @return this builder
      */
+    @SuppressWarnings("UnusedReturnValue")
```
reta, line 153: I don't understand why these suppression are needed?

lilyevsky: I put it because compiler gave me a warning about it. I can remove that suppression if you prefer.

reta: Please remove, thank you

lilyevsky: Done

lilyevsky: Hi @reta , please let me know what are the plans for merging.

flink-connector-opensearch/src/main/java/org/apache/flink/connector/opensearch/sink/OpensearchWriter.java
reta: @lilyevsky could you please address this and [this](https://github.com/apache/flink-connector-opensearch/pull/11/files#r1110066551) comment? thank you

lilyevsky: @reta No problem, could you please clarify: you want me to put back the IOException to both write and flush methods, correct? Also, I am not sure what you mean by addressing  [this](https://github.com/apache/flink-connector-opensearch/pull/11/files#r1110066551) . Is it about removing the "@SuppressWarnings("All")"?
Please confirm.

reta: @lilyevsky correct, the `throws IOException, InterruptedException` is the right part of the signature: the first comment is about `write`, the second is about `flush`, thank you

lilyevsky: @reta Done. Also removed the line with warnings suppressions that I added at some point.

reta (approved):

reta:
@MartijnVisser @zentol could you please take a look guys? thank you!

flink-connector-opensearch/src/main/java/org/apache/flink/connector/opensearch/sink/OpensearchSinkBuilder.java
reta: > Hi @reta , please let me know what are the plans for merging.

Thanks @lilyevsky , LGTM, asked Flink committers to take I look (I do not have commit rights)

lilyevsky: Thanks @reta

lilyevsky:
I see since I finalized this request there were some other changes in main branch, and so there are conflicts.
I am going to rebase my branch and hopefully resolve the conflicts this way.
After I do that, can we merge my request sooner rather than later? Otherwise there might be other conflicts again.

lilyevsky:
I actually resolved the conflicts in the web editor, and updated my branch. 
Can somebody review this?

reta:
@MartijnVisser @snuyanzin please help us here :-)

lilyevsky:
Actually, hold on for now. I was not able to compile it in my work environment, now tried in my home desktop and found some errors. Let me fix them. Will let you know.

lilyevsky:
I made it clean now, you can review and maybe merge already.
I got the latest from apache:main branch and re-applied all my changes on top of it, so you can see now it is just one commit.
I was able to locally build it.

reta (approved):

lilyevsky:
So you can merge it now?
After you do, I am going to build locally the latest version and do my own tests.

PR #14, opened 2023-03-31
[hotfix] Update flink, flink-shaded, archunit

snuyanzin:
The PR bumps dependencies

flink from 1.16.1 to 1.17.0
flink-shaded from 15.0 to 16.1 (same as in flink master)
~assertj from 3.21.0 to 3.24.2 (3.23.1 in flink master)~
archunit from 0.22.0 to 1.0.1 (1.0.0 in flink master)
junit5 from 5.8.1 to 5.9.2 (5.9.1 in flink master)
~testcontainers from 1.17.2 to 1.17.6 (same as in flink master)~
~mockito from 2.21.0 to 3.4.6 (same as in flink master)~

it also adds 
`archRule.failOnEmptyShould = false` since there are breaking changes in archunit 0.23.0 https://github.com/TNG/ArchUnit/releases/tag/v0.23.0
```
As mentioned in Enhancements/Core ArchRules will now by default reject evaluating if the set passed to the should-clause is empty. This will break existing rules that don't check any elements in their should-clause. You can restore the old behavior by setting the ArchUnit property archRule.failOnEmptyShould=false
```

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -63,10 +63,10 @@ under the License.
 		
 		<jackson-bom.version>2.13.4.20221013</jackson-bom.version>
 		<junit5.version>5.9.2</junit5.version>
-		<assertj.version>3.24.2</assertj.version>
+		<assertj.version>3.21.0</assertj.version>
 		<archunit.version>1.0.1</archunit.version>
-		<testcontainers.version>1.17.6</testcontainers.version>
-		<mockito.version>3.4.6</mockito.version>
+		<testcontainers.version>1.17.2</testcontainers.version>
+		<mockito.version>2.21.0</mockito.version>
 
 		<japicmp.skip>false</japicmp.skip>
 		<japicmp.referenceVersion>1.15.0</japicmp.referenceVersion>
--- a/pom.xml
+++ b/pom.xml
@@ -58,14 +58,15 @@ under the License.
 	</modules>
 
 	<properties>
-		<flink.version>1.16.1</flink.version>
-		<flink.shaded.version>15.0</flink.shaded.version>
+		<flink.version>1.17.0</flink.version>
+		<flink.shaded.version>16.1</flink.shaded.version>
 		
 		<jackson-bom.version>2.13.4.20221013</jackson-bom.version>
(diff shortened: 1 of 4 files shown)


pom.xml
```diff
@@ -58,12 +58,13 @@ under the License.
...
-		<flink.version>1.16.1</flink.version>
-		<flink.shaded.version>15.0</flink.shaded.version>
+		<flink.version>1.17.0</flink.version>
+		<flink.shaded.version>16.1</flink.shaded.version>
 		
 		<jackson-bom.version>2.13.4.20221013</jackson-bom.version>
-		<junit5.version>5.8.1</junit5.version>
+		<junit5.version>5.9.2</junit5.version>
 		<assertj.version>3.21.0</assertj.version>
+		<archunit.version>1.0.1</archunit.version>
```
MartijnVisser, line 67: You can actually completely remove this, it's not needed because there's no explicit usage of the archunit version in the repo anymore

MartijnVisser (approved):
@snuyanzin Thanks, I think we only need to remove the archunit.version from the POM.

reta:
@MartijnVisser the ArchUnit issue should be fixed, but it's very likely that we would need an update to 1.17 (this pull request) as well, the e2e tests would very likely fail. I am wondering if you or @dannycranmer could trigger Nightly GA manually to check that? (I have no permissions sadly). Thank you.

MartijnVisser:
> I am wondering if you or @dannycranmer could trigger Nightly GA manually to check that?

I've just done so at https://github.com/apache/flink-connector-opensearch/actions/runs/4611645665

> but it's very likely that we would need an update to 1.17 (this pull request) as well

Maybe. I've had to do a bunch of changes for other repos too, see for example RabbitMQ: https://github.com/apache/flink-connector-rabbitmq/commits/main

pom.xml
snuyanzin: thanks, makes sense

snuyanzin:
looks like nightly failure relates to archunt changes, should be fixed with archunit config change done within this PR

PR #15, opened 2023-04-03
[FLINK-31697] OpenSearch nightly CI failure

reta:
There was an issue with ArchUnit version: 1.16.1 uses 0.22.0 whereas 1.17.0 uses 1.0.0. The versions are not compatible so the suggestion is to drop explicit ArchUnit version management from connector and rely on `flink-architecture-tests-test` dependencies instead.


snuyanzin:
yeah, from 1.16 point of view probably makes sense, 
then with 1.17 update can update a bit configuration

PR #16, opened 2023-04-05
[hotfix] Fix nightly build for 1.18-SNAPSHOT, drop 1.16-SNAPSHOT from the nightly matrix

reta:
Fixing the compilation issue for 1.18 (see please https://github.com/apache/flink-connector-opensearch/actions/runs/4615625503/jobs/8159702829).

```
Error:  /home/runner/work/flink-connector-opensearch/flink-connector-opensearch/flink-connector-opensearch/src/test/java/org/apache/flink/connector/opensearch/OpensearchUtil.java:[144,19] org.apache.flink.connector.opensearch.OpensearchUtil.MockContext is not abstract and does not override abstract method getTargetColumns() in org.apache.flink.table.connector.sink.DynamicTableSink.Context
```

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -24,7 +24,7 @@ jobs:
     if: github.repository_owner == 'apache'
     strategy:
       matrix:
-        flink: [1.16.1, 1.17.0]
+        flink: [1.17.0]
     uses: apache/flink-connector-shared-utils/.github/workflows/ci.yml@ci_utils
     with:
       flink_version: ${{ matrix.flink }}
--- a/.github/workflows/weekly.yml
+++ b/.github/workflows/weekly.yml
@@ -26,7 +26,7 @@ jobs:
     if: github.repository_owner == 'apache'
     strategy:
       matrix:
-        flink: [1.16-SNAPSHOT, 1.17-SNAPSHOT, 1.18-SNAPSHOT]
+        flink: [1.17-SNAPSHOT, 1.18-SNAPSHOT]
     uses: apache/flink-connector-shared-utils/.github/workflows/ci.yml@ci_utils
     with:
       flink_version: ${{ matrix.flink }}
--- a/flink-connector-opensearch/src/test/java/org/apache/flink/connector/opensearch/OpensearchUtil.java
+++ b/flink-connector-opensearch/src/test/java/org/apache/flink/connector/opensearch/OpensearchUtil.java
@@ -43,6 +43,7 @@ import org.testcontainers.utility.DockerImageName;
 import java.security.KeyManagementException;
 import java.security.KeyStoreException;
 import java.security.NoSuchAlgorithmException;
+import java.util.Optional;
 
 /** Collection of utility methods for Opensearch tests. */
 @Internal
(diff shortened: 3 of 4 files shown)


.github/workflows/weekly.yml
```diff
@@ -26,7 +26,7 @@ jobs:
     if: github.repository_owner == 'apache'
     strategy:
       matrix:
-        flink: [1.16-SNAPSHOT, 1.17-SNAPSHOT, 1.18-SNAPSHOT]
```
reta, line 29: @MartijnVisser @snuyanzin I was not able to make the nightly e2e tests to run across all three versions (the changes in `ExecutionConfig` are resulting in NPEs all over the place for `1.16-SNAPSHOT`), I had to drop it.

snuyanzin: Looks reasonable.
The thing i would like to clarify whether it's ok to drop support for 1.16 while it's still maintained in main flink repo?
@MartijnVisser may be you have more info about this ?

MartijnVisser: > May I ask you please to brach off v3.0? I believe I have no permissions to do that

Sure. Do you want it to match the tag of the 3.0 release, or are there more commits that should be in there?

reta: I think d40b74ddba3cdaccfb45880aa0080d31dfa1e971 is good branching point (in case we would need to do another release), but it seem like the version should be 1.0.0 (not 3.0.0), and the main will become 2.0.0, is my understanding correct? Thank you.

MartijnVisser: @reta I've created [v1.0 ](https://github.com/apache/flink-connector-opensearch/tree/v1.0) and also set the version for `main` to `2.0-SNAPSHOT`.

reta: Thanks a lot @MartijnVisser , rebased, I think we are good to go

.github/workflows/release.yml
```diff
@@ -24,7 +24,7 @@ jobs:
     if: github.repository_owner == 'apache'
     strategy:
       matrix:
-        flink: [1.16.1, 1.17.0]
+        flink: [1.17.0]
```
MartijnVisser, line 27: I actually think you need to add `.github/workflows/push_pr.yml` to set the Flink version either to a matrix (with Flink 1.17.1 and potentially Flink 1.18-SNAPSHOT), or at least that version should be updated to 1.17.1 :)

reta: Moved it back to 1.17.0 ...

MartijnVisser (approved):
LGTM

reta:
> LGTM

Thanks a lot, may I ask you please to trigger nightly action manually when you have an opportunity, thank you (once it is green, I will replicate changes to Elasticsearch connector)

.github/workflows/release.yml
MartijnVisser: > I'll fix this afterwards :)

I'm an idiot and got confused by Flink 1.16.1, there's no Flink 1.17.1 yet 😅

MartijnVisser:
> may I ask you please to trigger nightly action manually

https://github.com/apache/flink-connector-opensearch/actions/runs/4669419298 is started!

.github/workflows/release.yml
reta: No worries at all

PR #17, opened 2023-04-10
[hotfix] Add opensearch.yml to enable flink to find the data

reswqa:
I compiled the flink documentation locally and found the problem shown in the figure below:

image

It seems that we used `sql_connector_download_table` shortcode in table document but did not introduce the corresponding yml file. As a result, we get an empty artifact.

### DIFF
--- a/docs/data/opensearch.yml
+++ b/docs/data/opensearch.yml
@@ -1,0 +1,21 @@
+################################################################################
+#  Licensed to the Apache Software Foundation (ASF) under one
+#  or more contributor license agreements.  See the NOTICE file
+#  distributed with this work for additional information
+#  regarding copyright ownership.  The ASF licenses this file
+#  to you under the Apache License, Version 2.0 (the
+#  "License"); you may not use this file except in compliance
+#  with the License.  You may obtain a copy of the License at
+#
+#      http://www.apache.org/licenses/LICENSE-2.0
+#
+#  Unless required by applicable law or agreed to in writing, software
+#  distributed under the License is distributed on an "AS IS" BASIS,
+#  WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+#  See the License for the specific language governing permissions and
+# limitations under the License.
+################################################################################
+
+variants:
+  - maven: flink-connector-opensearch
+    sql_url: https://repo.maven.apache.org/maven2/org/apache/flink/flink-sql-connector-opensearch/$full_version/flink-sql-connector-opensearch-$full_version.jar


MartijnVisser:
@reswqa We should not have specific docs branches. The release of 1.0.0 should use the docs from branch `v1.0`. That branch wasn't created until just now (which was missed somewhere), but I've just fixed that. I'll also sync your changes from `v1.0.0-docs` to `v1.0` and delete the dedicated docs branch.

MartijnVisser (approved):
LGTM

reswqa:
Thanks @MartijnVisser for the clarification. 
One thing I am not clear about is: in the `flink` repo, should we refer to the branch of the externalized connector(e.g. v3.0) or a specific release tag (e.g. v3.0.0)? By discussion with @dannycranmer  in FLINK-31704, I decided to adopt his suggestion to introduce a specific `release-tag-doc` branch to temporarily fix the documentation problem, and then delete this branch after the next release tag is generated. It seems that your proposal is that we directly track a specific branch.

MartijnVisser:
> It seems that your proposal is that we directly track a specific branch.

You either specify a tag or a branch. You can see it in the integration function here https://github.com/apache/flink/blob/master/docs/setup_docs.sh#L29-L31

The value for Opensearch is set at https://github.com/apache/flink/blob/master/docs/setup_docs.sh#L54, which currently is `v1.0.0` which gets set as the value for `ref`, so basically the value for the parameter `--branch` when doing the git clone. The `--branch` parameter accepts both branches, as well as tags. See https://git-scm.com/docs/git-clone
```
Instead of pointing the newly created HEAD to the branch pointed to by the cloned repository’s HEAD, point to <name> branch instead. In a non-bare repository, this is the branch that will be checked out. --branch can also take tags and detaches the HEAD at that commit in the resulting repository.
```

reswqa:
I know --branch can point to both branch names and tags. If I understand correctly, we only push tags like v3.0.0 forward during each release. So my main question is, if we modify a major bug in the document, do we have to wait for the next connector release so that users can see the correct version doc. If we use v1.0.0 in setup-docs.sh, will there be this problem?

MartijnVisser:
> So my main question is, if we modify a major bug in the document, do we have to wait for the next connector release so that users can see the correct version doc.

I don't think we should. I think we should update Flink to updates the references so they point to the branch, not the tag.

reswqa:
In fact, I prefer this approach, which is what I propose [here](https://issues.apache.org/jira/browse/FLINK-31704?focusedCommentId=17707954&page=com.atlassian.jira.plugin.system.issuetabpanels:comment-tabpanel#comment-17707954). But let's wait for @dannycranmer's opinion again. If we can reach an agreement, I am willing to create a ticket to do this.

dannycranmer:
It makes sense to use the `v3.0` branch instead of the dedicated docs branch. Assuming we are not adding new features for patch releases otherwise the docs may show features that are not actually released. @reswqa unless I misunderstood, I do not think it is the same as what you suggested. Since when we release `v3.1.0` there will be a `v3.1` branch and we still need to update the Flink repo. I understood your proposal to remove the requirement to update the branch in the Flink repo.

reswqa:
> unless I misunderstood, I do not think it is the same as what you suggested. Since when we release v3.1.0 there will be a v3.1 branch and we still need to update the Flink repo. I understood your proposal to remove the requirement to update the branch in the Flink repo.

Maybe I didn't fully express what I meant in our previous discuss. In my opinion, if the `v3.1` branch is created, we need to go to the flink repository to update it. 😆

reswqa:
Very thanks to @MartijnVisser and @dannycranmer for your suggestions and discussions. I have create FLINK-31779 to do this and then delete all dedicated doc branches.

PR #24, opened 2023-05-31
[FLINK-31856] Add support for Opensearch Connector REST client customization

reta:
It is not currently possible to customize the Opensearch REST client in all of the connectors.

### DIFF
--- a/flink-connector-opensearch/src/main/java/org/apache/flink/connector/opensearch/sink/RestClientFactory.java
+++ b/flink-connector-opensearch/src/main/java/org/apache/flink/connector/opensearch/sink/RestClientFactory.java
@@ -1,0 +1,104 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one or more
+ * contributor license agreements.  See the NOTICE file distributed with
+ * this work for additional information regarding copyright ownership.
+ * The ASF licenses this file to You under the Apache License, Version 2.0
+ * (the "License"); you may not use this file except in compliance with
+ * the License.  You may obtain a copy of the License at
+ *
+ *    http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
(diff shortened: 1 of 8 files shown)


flink-connector-opensearch/src/main/java/org/apache/flink/connector/opensearch/sink/RestClientFactory.java
```diff
@@ -0,0 +1,104 @@
...
+@PublicEvolving
+public interface RestClientFactory extends Serializable {
+
+    /** The REST client configuration. */
+    @PublicEvolving
+    interface RestClientConfig {
+        /**
+         * Gets the configured username.
+         *
+         * @return the configured username
```
snuyanzin, line 42: I wonder whether it makes sense to keep only `@return` part for all the methods since they are same...
WDYT?

snuyanzin:
Thanks @reta 
In general it looks ok

I have a minor comment

\+ in jira it was mentioned a case 
>We are currently using the using the OpensearchSink in [connector/opensearch/sink/OpensearchSink.java](https://github.com/apache/flink-connector-opensearch/blob/main/flink-connector-opensearch/src/main/java/org/apache/flink/connector/opensearch/sink/OpensearchSink.java) and need to be able to authenticate/authorise with Opensearch using OAuth2 and therefore need to be able to pass a bearer token to the bulk index calls.

It would be great to add it as ITTest, WDYT?

flink-connector-opensearch/src/main/java/org/apache/flink/connector/opensearch/sink/RestClientFactory.java
reta: I mean this is javadoc req :)

reta:
> It would be great to add it as ITTest, WDYT?

This functionality was always tested, the customization logic was inside single method, now it was moved into the class

snuyanzin (approved):
LGTM
thanks @reta

PR #50, opened 2025-04-11
[FLINK-37652][Connectors/Opensearch] Move to Flink to 2.0.0

reta:
[FLINK-37652][Connectors/Opensearch] Move to Flink to 2.0.0


reta:
@snuyanzin whenever you have time to look, please

.github/workflows/push_pr.yml
```diff
@@ -28,13 +28,8 @@ jobs:
...
       matrix:
-        flink: [ 1.18-SNAPSHOT ]
-        jdk: [ '8, 11, 17' ]
-        include:
-          - flink: 1.19-SNAPSHOT
-            jdk: '8, 11, 17, 21'
-          - flink: 1.20-SNAPSHOT
-            jdk: '8, 11, 17, 21'
+        flink: [ 2.1-SNAPSHOT ]
+        jdk: [ '11', '17, 21' ]
```
snuyanzin, line 32:
```suggestion
        jdk: [ '11, 17, 21' ]
```
is there a reason to isolate jdk11 in such a way?

.github/workflows/weekly.yml
```diff
@@ -30,23 +30,16 @@ jobs:
...
-            branch: main
-          }, 
+        }, 
         {
           flink: 1.20,
           jdk: '8, 11, 17, 21',
           branch: main
         }, {
-          flink: 1.19.0,
+          flink: 1.19.1,
```
snuyanzin, line 42: how about
```suggestion
          flink: 1.19.2,
```
?

pom.xml
```diff
@@ -380,28 +383,6 @@ under the License.
...
-				<jdk>[11,)</jdk>
-			</activation>
-			<modules>
-				<module>flink-connector-opensearch2</module>
-				<module>flink-connector-opensearch2-e2e-tests</module>
-				<module>flink-sql-connector-opensearch2</module>
-			</modules>
-		</profile>
 		<!-- The spotless comes from flink-connector-parent and needs version bump to support JDK-21 -->
 		<profile>
```
snuyanzin, line 387: may be put a comment here that we could enable it after dropping support of 1.20, since there is no issue with jdk11+

pom.xml
```diff
@@ -52,10 +52,13 @@ under the License.
...
+		<module>flink-connector-opensearch2-e2e-tests</module>
+		<module>flink-sql-connector-opensearch2</module>
 	</modules>
 
 	<properties>
-		<flink.version>1.20.0</flink.version>
+		<flink.version>2.0.0</flink.version>
 		<commons-compress.version>1.26.1</commons-compress.version>
 		<jackson-bom.version>2.16.2</jackson-bom.version>
 		<junit5.version>5.10.2</junit5.version>
```
snuyanzin, line 64:
```suggestion
		<jackson-bom.version>2.18.2</jackson-bom.version>
		<junit5.version>5.11.4</junit5.version>
```
how about bumping it to at least the same that is for flink 2.0?

pom.xml
```diff
@@ -64,7 +67,7 @@ under the License.
 		<mockito.version>3.12.4</mockito.version>
```
snuyanzin, line 67:
```suggestion
		<mockito.version>5.14.2</mockito.version>
```
as for flink 2.0, in case it is not that hard

pom.xml
```diff
@@ -64,7 +67,7 @@ under the License.
...
 
 		<japicmp.skip>false</japicmp.skip>
-		<japicmp.referenceVersion>1.0.0-1.19</japicmp.referenceVersion>
+		<japicmp.referenceVersion>1.0.0-1.20</japicmp.referenceVersion>
 
 		<slf4j.version>1.7.36</slf4j.version>
 		<log4j.version>2.17.2</log4j.version>
```
snuyanzin, line 73:
```suggestion
		<log4j.version>2.24.1</log4j.version>
```
as for 2.0

snuyanzin (commented):
Great to see support of 2.0 coming
I left some comments

.github/workflows/push_pr.yml
reta: No, not really, fixing it, thanks!

reta:
> I left some comments

Thanks a lot, @snuyanzin ! Addressed all except `spotless` version bump (I think we should do it in flink-connector-parent), thank you!

.github/workflows/weekly.yml
```diff
@@ -30,23 +30,16 @@ jobs:
...
         {
           flink: 1.20,
           jdk: '8, 11, 17, 21',
           branch: main
         }, {
-          flink: 1.19.0,
+          flink: 1.19.2,
           jdk: '8, 11, 17, 21',
           branch: v1.2
         }, {
```
snuyanzin, line 45: sorry, I think I missed it:
since 1.18-SNAPSHOT was removed in this PR
how about replacing 1.18.2 (which is a couple of lines below) with 1.20.1 ?

pom.xml
snuyanzin: makes sense, probably need to enable spotless for jdk21 there
i'll submit a PR there

snuyanzin:
thanks for addressing feedback 👍 ,
 I think there is one more minor and we are good to merge

.github/workflows/weekly.yml
reta: We already run against 1.20 (https://github.com/apache/flink-connector-opensearch/pull/50/files#diff-b077dee47bccc37c2b6fef32cdd102f36acadad73ede1f410c3941f39c50cd94R38), do you mean replacing `1.20` with `1.20.1`?

snuyanzin (approved):

snuyanzin:
yep, thanks for addressing feedback and sorry for the delay
Will do that by the end of the week since now in a conf without a right laptop for this 🙈

PR #62, opened 2026-05-12
[hotfix] Make PackagingITCase Windows-friendly

reta:
Make PackagingITCase Windows-friendly
Also, added PackagingITCase for `flink-sql-connector-opensearch3`

### DIFF
--- a/flink-sql-connector-opensearch2/src/test/java/org/apache/flink/connector/opensearch2/PackagingITCase.java
+++ b/flink-sql-connector-opensearch2/src/test/java/org/apache/flink/connector/opensearch2/PackagingITCase.java
@@ -32,7 +32,8 @@ class PackagingITCase {
     @Test
     void testPackaging() throws Exception {
         final Path jar =
-                ResourceTestUtils.getResource(".*/flink-sql-connector-opensearch2-[^/]*\\.jar");
+                ResourceTestUtils.getResource(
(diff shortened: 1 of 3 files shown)


flink-sql-connector-opensearch/src/test/java/org/apache/flink/connector/opensearch/PackagingITCase.java
```diff
@@ -32,7 +32,7 @@ class PackagingITCase {
     @Test
     void testPackaging() throws Exception {
         final Path jar =
-                ResourceTestUtils.getResource(".*/flink-sql-connector-opensearch-[^/]*\\.jar");
+                ResourceTestUtils.getResource(".*[\\\\/]flink-sql-connector-opensearch-[^\\\\/]*\\.jar");
```
reta: Windows uses `\` as path separator

reta:
@snuyanzin could you please take a look when you have a minute? thank you

snuyanzin:
should we have windows in ci may for only one jdk version?

reta:
> should we have windows in ci may for only one jdk version?

OK, so this is not as easy I thought, the CI flows are coming from `apache/flink-connector-shared-utils/.github/workflows/ci.yml` and it only has `ubuntu` runners, we may need to changed that one first

snuyanzin (approved):

PR #76, opened 2026-09-18
[hotfix] [opensearch] Fix weekly Github workflow

reta:
Fix weekly Github workflow


.github/workflows/weekly.yml
```diff
@@ -36,13 +36,13 @@ jobs:
...
-          jdk: '11, 17, 21',
-          branch: main
+          jdk: '8, 11, 17, 21',
+          branch: v1.2
         }, 
         {
           flink: 1.20.5,
           jdk: '8, 11, 17, 21',
-          branch: main
+          branch: v1.2
```
snuyanzin, line 45: should we also drop 1.18 and 1.19 as no supported?

reta: I think we could drop 1.18, but other connectors (https://github.com/apache/flink-connector-elasticsearch/blob/main/.github/workflows/weekly.yml) claim 1.19 support still

reta: Dropped 1.18

snuyanzin (approved):
LGTM

PR #77, opened 2026-09-18
[FLINK-40711][ci] Do not run weekly if there is no new commits since last green run

snuyanzin:
The PR is similar https://github.com/apache/flink/pull/28348
which was done for Flink main repo
since Apache Infra sometimes complains about our resource usages

at the same time it is still possible to force run it manually

example of. Apache Infra complains
https://lists.apache.org/thread/jn9t7qjj4wlz98t1ddqxwypcn6g7wrds


.github/workflows/weekly-trigger.yml
```diff
@@ -0,0 +1,99 @@
...
+              await trigger('no previous weekly run found');
+              return;
+            }
+            if (lastConclusion !== 'success') {
+              await trigger(`last weekly run concluded '${lastConclusion ?? 'unknown'}', re-running`);
+              return;
+            }
+
+            // Last weekly was green: skip unless any watched branch has new commits since it started (manual runs still work).
+            for (const branch of branches) {
```
reta: So the weekly runs against Flink snapshots as well, so that means if there are new commits in Flink (that could potentially impact connector), we will not run weekly if there are no commits in branch(es)?

snuyanzin: yeah, good catch
I splitted into 2 parts: snapshot and fixed versions to mitigate it

reta (approved):

PR #78, opened 2026-09-20
[hotfix] [opensearch] Update to Apache Flink 2.3.0 baseline

reta:
Update to Apache Flink 2.3.0 baseline


flink-connector-opensearch2/src/test/java/org/apache/flink/connector/opensearch/table/Opensearch2DynamicSinkITCase.java
```diff
@@ -163,7 +163,7 @@ public void testWritingDocuments() throws Exception {
                         .getSource();
         Map<Object, Object> expectedMap = new HashMap<>();
         expectedMap.put("a", 1);
-        expectedMap.put("b", "00:00:12");
+        expectedMap.put("b", "00:00:12.135");
```
snuyanzin: yeah, I remember this =)

for the history: result of fix within https://issues.apache.org/jira/browse/FLINK-17224

pom.xml
```diff
@@ -55,7 +55,7 @@ under the License.
 	</modules>
 
 	<properties>
-		<flink.version>2.0.0</flink.version>
+		<flink.version>2.3.0</flink.version>
 		<commons-compress.version>1.26.1</commons-compress.version>
 		<jackson-bom.version>2.18.2</jackson-bom.version>
```
snuyanzin: should we also bump other versions?
for example Flink 2.3.x depends on 2.20.1 jackson
https://github.com/apache/flink/blob/ae68dd4950c1485fc470cb02c08cc02b59ce20aa/pom.xml#L154

reta: Will do, thanks!

snuyanzin (approved):

